PR #15, opened 2014-03-10
Adding Multipart Decoder, redux

mythguided:
Going with rename as first commit as requested.


sigmavirus24:
@mythguided I realized you never added yourself to [AUTHORS](https://github.com/sigmavirus24/requests-toolbelt/blob/master/AUTHORS.rst), a pull request doing so would be more than welcome.

PR #21, opened 2014-04-14
Add AuthHandler feature

sigmavirus24:
TODO:

- [x] Add some integration tests
- [x] Decide whether to allow for "domains" like `*.github.com`


sigmavirus24:
I keep wavering on whether or not this pull request is a good idea. @Lukasa thoughts?

Lukasa:
I think it is. =)

sigmavirus24:
I got nervous about not checking URL schemes, so I added that as part of the strategy key as well as adding tests for it. Once the tests pass, I'll merge this.

PR #40, opened 2014-09-01
import from urllib3 directly if it has not been shipped

jbremer:
https://github.com/untitaker/vdirsyncer/issues/82#issuecomment-47888471

Surprised this hasn't been fixed already ;)


sigmavirus24:
This is no longer necessary in recent versions of pip.

sigmavirus24:
Thanks @jbremer

PR #61, opened 2015-03-03
Initial implementation of fingerprint adapter.

Lukasa:
This is a transport adapter that allows assertion of a specific certificate fingerprint, rather than simply asserting a hostname match on a certificate. This allows certificate pinning in requests by ensuring that only one certificate can possibly be matched on a given connection.

There's scope for additional work here. For example, it might be helpful to allow a callback to be called to provide the fingerprint, though that will require co-operation with urllib3 if it's believed to be useful.

I'm also currently missing tests, though hopefully I'll add them shortly.


sigmavirus24:
ಠ_ಠ pep8 1.6 (╯°□°)╯︵ ┻━┻

requests_toolbelt/__init__.py
```diff
@@ -16,7 +16,7 @@
 __version__ = '0.3.1'
 __version_info__ = tuple(int(i) for i in __version__.split('.'))
 
-from .adapters import SSLAdapter, SourceAddressAdapter
+from .adapters import SSLAdapter, SourceAddressAdapter, FingerprintAdapter
```
sigmavirus24: Hadn't we discussed discontinuing this practice?

requests_toolbelt/adapters/fingerprint.py
```diff
@@ -0,0 +1,44 @@
...
+    def __init__(self, fingerprint, **kwargs):
+        self.fingerprint = fingerprint
+
+        super(FingerprintAdapter, self).__init__(**kwargs)
+
+    def init_poolmanager(self, connections, maxsize, block=False):
+        self.poolmanager = PoolManager(num_pools=connections,
+                                       maxsize=maxsize,
+                                       block=block,
+                                       assert_fingerprint=self.fingerprint)
```
sigmavirus24: Didn't @t-8ch just add support for other kinds of fingerprints or are they all the same?

tests/test_fingerprintadapter.py
```diff
@@ -0,0 +1,20 @@
...
+class TestFingerprintAdapter(unittest.TestCase):
+    HTTP2BIN_FINGERPRINT = 'abf8683eeba8521ad2e8dc48e92a1cbea3ff8608f1417948fdad75d7b50eb264'
+
+    def setUp(self):
+        self.session = requests.Session()
+        self.session.mount('https://http2bin.org', FingerprintAdapter(self.HTTP2BIN_FINGERPRINT))
+        self.recorder = get_betamax(self.session)
+
+    def test_klevas(self):
+        with self.recorder.use_cassette('http2bin_fingerprint'):
```
sigmavirus24: ❤︎

t-8ch:
It just looks at the length of the fingerprint

tests/test_fingerprintadapter.py
```diff
@@ -0,0 +1,20 @@
...
+
+class TestFingerprintAdapter(unittest.TestCase):
+    HTTP2BIN_FINGERPRINT = 'abf8683eeba8521ad2e8dc48e92a1cbea3ff8608f1417948fdad75d7b50eb264'
+
+    def setUp(self):
+        self.session = requests.Session()
+        self.session.mount('https://http2bin.org', FingerprintAdapter(self.HTTP2BIN_FINGERPRINT))
+        self.recorder = get_betamax(self.session)
+
+    def test_klevas(self):
```
sigmavirus24: klevas? Is this copied from ssladapter?

requests_toolbelt/__init__.py
Lukasa: Not that I recall. What would you like instead? =)

requests_toolbelt/adapters/fingerprint.py
Lukasa: Not that I can see in urllib3's code...

tests/test_fingerprintadapter.py
Lukasa: Ugh, yes, good catch.

requests_toolbelt/adapters/fingerprint.py
sigmavirus24: They're all the same parameter as he kindly pointed out. They're distinguished by their length though.

requests_toolbelt/__init__.py
sigmavirus24: We kind of discussed it here: https://github.com/sigmavirus24/requests-toolbelt/issues/54

I think we should start trending towards what we discussed in that issue. I know people will open issues regarding it, but I need to update the documentation either way.

Lukasa:
Should have fixed the outstanding stuff in 987f0e34a48c67e3378751223f57d0cf0c56f481.

sigmavirus24:
Rebased your change @Lukasa . Once it passes (again) I'll merge it. :cake:

PR #64, opened 2015-03-06
Naive Session Thread Pool

sigmavirus24:
This is entirely untested (even by hand). Use at your own risk

TODO:

- [x] Test with `stream=True`
- [x] Write tests
- [x] Document behaviour


sigmavirus24:
So @Lukasa this works. You can toy around with it using `python -i examples/threading/threaded.py`. I'm surprised exactly how little work it took.

We just need to put together a good API for it. I think mimicking `multiprocessing` is a good start as a matter of familiarity. Thoughts?

Lukasa:
I think that's a good start. It'd be very interesting to consider how many other useful APIs there might be to it. For example, what about a coroutine-based one? `send()` to make requests, `next()` to receive responses. Not sure how much work it'd be to write, but it feels like a fun paradigm.

sigmavirus24:
Using `stream=True` seems to Just Work™. The docs, in their current form, probably could use some rewriting. That said, they're a start. I'd like more examples with more features. Appropriate caveats about how timeouts affect the performance (for example if you run `python examples/threaded/threading.py` it will hang for 5 seconds as the request with a 5 second timeout completes). Finally, this still needs tests.

I'd love to have a coroutine like API for this as well, but I'm not (yet) familiar enough with asyncio to author that myself. Pull requests accepted of course :).

PR #78, opened 2015-03-29
Allow files as input for StreamingIterator

### LINKED ISSUE #38: Add from_file classmethod to StreamingIterator
This allows someone to pass something like `sys.stdin` to `StreamingIterator.from_file` without having to create a function to wrap it. We can handle that for them.

untitaker:
Fixes #38

The usecase for this is unsized file-like objects like sys.stdin.

This functionality could've been implemented as a classmethod or helper
function. However, when using StreamingIterator with file-like objects,
it would've been quite inefficient to

1.) Read data from the underlying file at a fixed chunk size for the iterator
2.) Then maintain a buffer for the file-like interface provided on top
    of that iterator

### DIFF
--- a/requests_toolbelt/streaming_iterator.py
+++ b/requests_toolbelt/streaming_iterator.py
@@ -42,8 +42,18 @@ class StreamingIterator(object):
 
         r = requests.post(url, data=StreamingIterator(size, iterator))
 
+    You can also pass file-like objects to :py:class:`StreamingIterator` in
+    case requests can't determize the filesize itself. This is the case with
+    streaming file objects like ``stdin`` or any sockets. **Wrapping e.g. files
+    that are on disk with ``StreamingIterator`` is unnecessary**, because
+    requests can determize the filesize itself.
+
     Naturally, you should also set the `Content-Type` of your upload
     appropriately because the toolbelt will not attempt to guess that for you.
+
+    .. versionchanged:: 0.4
+
+        Files are accepted as input.
 
     """
 
--- a/requests_toolbelt/streaming_iterator.py
+++ b/requests_toolbelt/streaming_iterator.py
@@ -60,6 +60,23 @@ class StreamingIterator(object):
         #: body. See bug #80 for more details
         self.len = self.size
 
+        #: Encoding the input data is using
+        self.encoding = encoding
+
+        #: The iterator used to generate the upload data
+        self.iterator = iterator
+
+        if hasattr(iterator, 'read'):
+            self._file = iterator
+        else:
+            self._file = _IteratorAsBinaryFile(iterator, encoding)
+
+    def read(self, size=-1):
+        return encode_with(self._file.read(size), self.encoding)
+
+
+class _IteratorAsBinaryFile(object):
+    def __init__(self, iterator, encoding='utf-8'):
         #: The iterator used to generate the upload data
         self.iterator = iterator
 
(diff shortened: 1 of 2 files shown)


untitaker:
@sigmavirus24 I hope this gets my point better across than my babbling about rewriting internal APIs.

sigmavirus24:
I can't look at this right now but I appreciate the effort. I should be able to look in a few hours

untitaker:
Ok. If you like the concept I'll add tests, changelog etc.

sigmavirus24:
So this looks okay but there are backwards incompatible changes here. `iterator` is a public attribute of `StreamingIterator`. No one *should* be modifying it, but if they are this will break them as is. Further the parameter name was previously `iterator` and anyone using a keyword argument should be able to continue doing so.

untitaker:
* What should `self.iterator` be in case a file was passed?
* Should files be passed through the same argument?

sigmavirus24:
> What should self.iterator be in case a file was passed?

Exactly what it is in this pull request (instead it's just not called "file").

> Should files be passed through the same argument?

Yes. Most (if not all) files are iterable.

untitaker:
>Most (if not all) files are iterable.

Didn't think about it that way.

untitaker:
Applied your suggestions.

untitaker:
Should we hint in the docs that passing `f` instead of `iter(f)` might yield performance benefits if `not hasattr(iter(f), 'read')`?

sigmavirus24:
I think we should more simply state in the docs that generator objects as well as open file objects providing a `read` method, can both be passed in. It should be stressed that this should only be necessary for streaming file objects like `sys.stdin` or a `socket`, which the user might know the length of but requests cannot determine for them. For files that are on disk, requests can determine that length.

untitaker:
I added the necessary documentation. Should I create a new changelog section for 0.4 as well?

untitaker:
Regarding tests: Is the testsuite intentionally kept close to the unittest style (so it's runnable with e.g. nose) or may I use more exotic pytest-specific features?

sigmavirus24:
> Regarding tests: Is the testsuite intentionally kept close to the unittest style (so it's runnable with e.g. nose) or may I use more exotic pytest-specific features?

No and yes.

untitaker:
Added tests. I used [py.test fixtures](https://pytest.org/latest/fixture.html#fixture-parametrize), for reference.

sigmavirus24:
@untitaker do you mind rebasing your PR?

requests_toolbelt/streaming_iterator.py
```diff
@@ -41,9 +41,18 @@ class StreamingIterator(object):
...
+    case requests can't determize the filesize itself. This is the case with
+    streaming file objects like ``stdin`` or any sockets. **Wrapping e.g. files
+    that are on disk with ``StreamingIterator`` is unnecessary**, because
+    requests can determize the filesize itself.
+
     Naturally, you should also set the `Content-Type` of your upload
     appropriately because the toolbelt will not attempt to guess that for you.
 
+    .. versionchanged:: 0.4
+        Files are accepted as input.
```
sigmavirus24: Not to be a jerk, but this won't properly render unless there's an extra new-line between the directive and the comment.

untitaker:
Rebased and fixed the formatting issue.

sigmavirus24:
Thanks @untitaker! :cake:

PR #81, opened 2015-04-01
Use a len attribute instead of the __len__ method

### LINKED ISSUE #80: Can't encode files larger than 4GB on 32-bit system.
Using Python 3.4.0 and Requests 2.6.0

```python
    content = MultipartEncoder(fields=[('metadata', ("", json.dumps(metadata))), ('content', (file_name, open(file_path, 'rb'), mime_type))])
```

Then, I call requests.post and get this error.

```python
File "/home/xxxx/upload.py", line 222, in upload
    r = requests.post(url, headers=headers, data=content)
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests/api.py", line 108, in post
    return request('post', url, data=data, json=json, **kwargs)
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests/api.py", line 50, in request
    response = session.request(method=method, url=url, **kwargs)
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests/sessions.py", line 443, in request
    data = data or {},
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests_toolbelt/multipart/encoder.py", line 96, in __len__
    return self._len or self._calculate_length()
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests_toolbelt/multipart/encoder.py", line 111, in _calculate_length
    ) + boundary_len + 4
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests_toolbelt/multipart/encoder.py", line 110, in <genexpr>
    (boundary_len + len(p) + 4) for p in self.parts
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests_toolbelt/multipart/encoder.py", line 377, in __len__
    return len(self.headers) + super_len(self.body)
  File "/home/xxxx/upload/venv/lib/python3.4/site-packages/requests/utils.py", line 52, in super_len
    return len(o)
OverflowError: cannot fit 'int' into an index-sized integer
```

`__len__` can't return values larger than a C int.
http://bugs.python.org/issue21444

sigmavirus24:
On 32-bit architecture, the size of the integer returned by the __len__
method are restricted to the maximum size of an integer in C. Any number
that is larger than 2**32 results in an OverflowError.

requests uses a utility method to determine the length of a body of a
request. By setting `len` as an attribute, we can still take advantage of
super_len's behaviour and we can hack around the OverflowError.

Closes #80

### DIFF
--- a/requests_toolbelt/multipart/encoder.py
+++ b/requests_toolbelt/multipart/encoder.py
@@ -122,7 +122,23 @@ class MultipartEncoder(object):
         # Load boundary into buffer
         self._write_boundary()
 
-    def __len__(self):
+    @property
+    def len(self):
+        """Length of the multipart/form-data body.
+
+        requests will first attempt to get the length of the body by calling
+        ``len(body)`` and then by checking for the ``len`` attribute.
+
+        On 32-bit systems, the ``__len__`` method cannot return anything
+        larger than an integer (in C) can hold. If the total size of the body
+        is even slightly larger than 4GB users will see an OverflowError. This
+        manifested itself in `bug #80`_.
+
+        As such, we now calculate the length lazily as a property.
+
+        .. _bug #80:
+            https://github.com/sigmavirus24/requests-toolbelt/issues/80
+        """
         # If _len isn't already calculated, calculate, return, and set it
         return self._len or self._calculate_length()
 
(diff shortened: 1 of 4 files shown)


Lukasa:
Well, that makes me sad. =(

sigmavirus24:
Me too. While trying to debug this last night on IRC with @Lyrrad we found http://bugs.python.org/issue21444 which makes it seem like something they won't fix. They didn't even fix it in Python 3. It makes me wonder exactly how tenable that restriction really is, especially considering the expansion of Python into data sciences.

sigmavirus24:
Btw, @untitaker you may need to rebase your PR after this merges.

PR #87, opened 2015-06-05
tee module

sigmavirus24:
This adds the tee module with two functions for streaming a response and writing it to a file-like object as well as yielding the chunk.

TODO:

- [x] Add tests
- [x] Add documentation

---

I picked the largest download from CirrOS (8+ MB) and tested it out. Given the simplicity, it would be hard for this to fail :D.

```py
>>> import requests
>>> from requests_toolbelt.downloadutils import tee as t
>>> r = requests.get('https://download.cirros-cloud.net/0.3.4/cirros-0.3.4-x86_64-uec.tar.gz', stream=True, verify=False)
>>> fd = open('cirros-0.3.4-x86_64-uec.tar.gz', 'wb')
>>> sum(len(c) for c in t.tee(r, fd))
8683894
>>> r.headers['content-length']
'8683894'
>>> fd.close()
```

### DIFF
--- a/docs/downloadutils.rst
+++ b/docs/downloadutils.rst
@@ -5,3 +5,9 @@ Utilities for Downloading Streaming Reponses
 
 .. autofunction::
     requests_toolbelt.downloadutils.stream.stream_response_to_file
+
+.. autofunction::
+    requests_toolbelt.downloadutils.tee.tee
+
+.. autofunction::
+    requests_toolbelt.downloadutils.tee.tee_to_file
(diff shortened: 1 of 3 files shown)


sigmavirus24:
Also, can I just point out that it's a travesty that https://download.cirros-cloud.net uses a self-signed certificate. Tons of people download those files per day to test OpenStack and most don't have a copy of the certificate to verify with, so by ignoring that and running the images, they could just get some hijacked copy and bork everything. =(

PR #90, opened 2015-06-13
Add dump utils for request-response cycle dumping

### LINKED ISSUE #89: We should have a debug-dump mode that dumps requests/responses in their wire form

sigmavirus24:
Closes #89 

---

TODO:

- [x] Tests
- [x] Docs

---

Test script:

```py
import requests

from requests_toolbelt.utils import dump


getresp = requests.get('https://httpbin.org/get', params={'biz': 'baz'})
postresp = requests.post('https://httpbin.org/post', json={'foo': 'bar'})

getdata = dump.dump_response(getresp)
postdata = dump.dump_response(postresp)

print(getdata.decode())
print(postdata.decode())
```

Outputs:

```
< GET /get HTTP/1.1
< Host: httpbin.org
< Accept: */*
< Accept-Encoding: gzip, deflate
< Connection: keep-alive
< User-Agent: python-requests/2.7.0 CPython/2.7.9 Darwin/14.1.0
<

> HTTP/1.1 200 OK
> access-control-allow-credentials: true
> access-control-allow-origin: *
> connection: keep-alive
> content-length: 293
> content-type: application/json
> date: Sat, 13 Jun 2015 17:38:22 GMT
> server: nginx
>
{
  "args": {
    "biz": "baz"
  },
  "headers": {
    "Accept": "*/*",
    "Accept-Encoding": "gzip, deflate",
    "Host": "httpbin.org",
    "User-Agent": "python-requests/2.7.0 CPython/2.7.9 Darwin/14.1.0"
  },
  "origin": "69.29.227.64",
  "url": "https://httpbin.org/get?biz=baz"
}

< POST /post HTTP/1.1
< Host: httpbin.org
< Accept: */*
< Accept-Encoding: gzip, deflate
< Connection: keep-alive
< Content-Length: 14
< Content-Type: application/json
< User-Agent: python-requests/2.7.0 CPython/2.7.9 Darwin/14.1.0
<
< {"foo": "bar"}
> HTTP/1.1 200 OK
> access-control-allow-credentials: true
> access-control-allow-origin: *
> connection: keep-alive
> content-length: 435
> content-type: application/json
> date: Sat, 13 Jun 2015 17:38:22 GMT
> server: nginx
>
{
  "args": {},
  "data": "{\"foo\": \"bar\"}",
  "files": {},
  "form": {},
  "headers": {
    "Accept": "*/*",
    "Accept-Encoding": "gzip, deflate",
    "Content-Length": "14",
    "Content-Type": "application/json",
    "Host": "httpbin.org",
    "User-Agent": "python-requests/2.7.0 CPython/2.7.9 Darwin/14.1.0"
  },
  "json": {
    "foo": "bar"
  },
  "origin": "69.29.227.64",
  "url": "https://httpbin.org/post"
}
```

I'd love some preliminary review @Lukasa if you have some time.


sigmavirus24:
Also, since @kennethreitz was interested in this too, I'd love preliminary review from you too.

requests_toolbelt/utils/dump.py
```diff
@@ -0,0 +1,166 @@
+"""This module provides functions for dumping information about responses."""
+import collections
+
+from requests import compat
+
+
+__all__ = ('dump_response',)
+_PrefixSettings = collections.namedtuple('PrefixSettings',
+                                         ['request', 'response'])
```
Lukasa: I know this is weird, but I'd rather have this declaration closer to its subclass. I just don't love having them separated by an unrelated block of code.

sigmavirus24: That's fair

requests_toolbelt/utils/dump.py
```diff
@@ -0,0 +1,166 @@
...
+    prefix = prefixes.request
+    method = _coerce_to_bytes(request.method)
+    uri = compat.urlparse(request.url)
+
+    request_path = _coerce_to_bytes(uri.path)
+    if uri.query:
+        request_path + b'?' + _coerce_to_bytes(uri.query)
+
+    # <prefix><METHOD> <request-path> HTTP/1.1
+    bytearr.extend(prefix + method + b' ' + request_path + b' HTTP/1.1\r\n')
```
Lukasa: So this is a lie if we contact a proxy using plaintext HTTP. Is there any way we can check that from here?

requests_toolbelt/utils/dump.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # <prefix>Host: <request-host> OR host header specified by user
+    headers = request.headers.copy()
+    host_header = _coerce_to_bytes(headers.pop('Host',
+                                   _coerce_to_bytes(uri.netloc)))
```
Lukasa: Does this need to go through `_coerce_to_bytes` twice?

requests_toolbelt/utils/dump.py
```diff
@@ -0,0 +1,166 @@
...
+
+    # <prefix>HTTP/<version_str> <status_code> <reason>
+    bytearr.extend(prefix + b'HTTP/' + version_str + b' ' +
+                   str(raw.status).encode() + b' ' +
+                   _coerce_to_bytes(raw.reason) + b'\r\n')
+
+    headers = raw.headers
+    for name in sorted(headers.keys()):
+        for value in headers.getlist(name):
+            bytearr.extend(prefix + _format_header(name, value))
```
Lukasa: I think reaching into httplib is a better idea here. With responses we can basically work out exactly what they looked like, so we should do that. This sorting of header names is really less than ideal.

requests_toolbelt/utils/dump.py
```diff
@@ -0,0 +1,166 @@
...
+    # Let's interact almost entirely with urllib3's response
+    raw = response.raw
+
+    # Let's convert the version int from httplib to bytes
+    version_str = HTTP_VERSIONS.get(raw.version, b'?')
+
+    # <prefix>HTTP/<version_str> <status_code> <reason>
+    bytearr.extend(prefix + b'HTTP/' + version_str + b' ' +
+                   str(raw.status).encode() + b' ' +
+                   _coerce_to_bytes(raw.reason) + b'\r\n')
```
Lukasa: Best not to use a naked `encode()` here, I worry about what might happen with non-ascii-compatible default encodings. Better to say `'ascii'` explicitly.

requests_toolbelt/utils/dump.py
```diff
@@ -0,0 +1,166 @@
...
+    :param response_prefix: (*optional*)
+        Bytes to prefix each line of the response data
+    :type response_prefix: :class:`bytes`
+    :param data_array: (*optional*)
+        Bytearray to which we append the request-response cycle data
+    :type data_array: :class:`bytearray`
+    :returns: Formatted bytes of request and response information.
+    :rtype: :class:`bytearray`
+    """
+    data = data_array or bytearray()
```
Lukasa: `data = data_array if data_array is not None else bytearray()` feels better to me, but that's a minor point.

Lukasa:
Some review comments inline, but the general shape of this is great. =D

requests_toolbelt/utils/dump.py
sigmavirus24: Nope. Responses never contain information about that stuff. urllib3's HTTPResponse should have a ref to the connection which we might be able to introspect though. I'll have to toy with that.

requests_toolbelt/utils/dump.py
sigmavirus24: Nope.

requests_toolbelt/utils/dump.py
Lukasa: Oooh, that's a good idea, that should work fine. Look for things with the name like `tunnel` or `_tunnel`.

sigmavirus24: So unless `stream=True` the best bet seems to be looking at `response.connection.proxy_manager` (`r.connection` is the `HTTPAdapter` that sent the request/generated the response). So what does the format look like for this?

FWIW, if I do:

```py
>>> r = requests.get('http://httpbin.org/get', proxies={'http': 'http://localhost:8088'})
>>> r.connection.proxy_manager
{'http://localhost:8088': <requests.packages.urllib3.poolmanager.ProxyManager object at 0x10ba94f10>}
```

(That's assuming mitmproxy running on port 8088)

requests_toolbelt/utils/dump.py
sigmavirus24: So `r.raw._fp.getheaders()` and `r.raw.getheaders()` return the same ordering of headers. I've removed the call to `sorted()`.

requests_toolbelt/utils/dump.py
sigmavirus24: Done.

requests_toolbelt/utils/dump.py
sigmavirus24: Done.

requests_toolbelt/utils/dump.py
sigmavirus24: Done.

requests_toolbelt/utils/dump.py
sigmavirus24: Done.

requests_toolbelt/utils/dump.py
Lukasa: The key thing to check is the scheme of the request. If the scheme is `https://` then the CONNECT verb was used on the original request and we can probably just elide the whole proxy request, so everything looks normal. If the scheme is `http://` then the request line changes to `GET http://httpbin.org/get HTTP/1.1`

sigmavirus24:
So detecting proxies will become harder since we now allow people to specify per-host proxies (https://github.com/kennethreitz/requests/pull/2741). Shouldn't be *that* much harder though.

sigmavirus24:
I rebased this and added work to capture the information around proxies. I still need to write tests for all of this, but it shouldn't be very difficult to do.

Also, fwiw, I didn't test any of the proxy information so if someone else wants to test it, that'd be great.

PR #97, opened 2015-08-09
Allow string types for source address adapter.

### LINKED ISSUE #96: SourceAddressAdapter requires IP + port tuple
accidentally submitted this as shazow/urllib3#691

I think in addition to accepting a tuple, if just the IP string is passed then toolbelt should use a random available port so the user doesn't have to do all the guesswork :)

Lukasa:
Resolves #96.

I'm a bit sad that this adapter has no test coverage, but it's a massive PITA to test effectively. I can write tests that are platform specific, but doing anything else is a huge pain. =(

### DIFF
--- a/requests_toolbelt/adapters/source.py
+++ b/requests_toolbelt/adapters/source.py
@@ -8,7 +8,7 @@ demonstrated on the Requests GitHub page.
 """
 from requests.adapters import HTTPAdapter
 
-from .._compat import poolmanager
+from .._compat import poolmanager, basestring
 
 
 class SourceAddressAdapter(HTTPAdapter):
--- a/requests_toolbelt/adapters/source.py
+++ b/requests_toolbelt/adapters/source.py
@@ -30,8 +30,12 @@ class SourceAddressAdapter(HTTPAdapter):
     def __init__(self, source_address, **kwargs):
         if isinstance(source_address, basestring):
             self.source_address = (source_address, 0)
-        else:
+        elif isinstance(source_address, tuple):
             self.source_address = source_address
+        else:
+            raise TypeError(
+                "source_address must be IP address string or (ip, port) tuple"
+            )
 
         super(SourceAddressAdapter, self).__init__(**kwargs)
 
--- a/requests_toolbelt/adapters/source.py
+++ b/requests_toolbelt/adapters/source.py
@@ -28,7 +28,10 @@ class SourceAddressAdapter(HTTPAdapter):
         s.mount('http://', SourceAddressAdapter('10.10.10.10'))
     """
     def __init__(self, source_address, **kwargs):
-        self.source_address = source_address
+        if isinstance(source_address, basestring):
+            self.source_address = (source_address, 0)
+        else:
+            self.source_address = source_address
 
         super(SourceAddressAdapter, self).__init__(**kwargs)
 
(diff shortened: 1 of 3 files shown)


sigmavirus24:
`basestring` doesn't exist on Python 3. Requests has a shim for that though if you'd like to add it to our `_compat` module.

Lukasa:
Oh I hate Python 3. Why on earth was that removed? Anyway, I'll update.

Lukasa:
Now flake8 is mad. =(

Lukasa:
I think realistically we should just exclude the `_compat.py` file from flake8 checking, because it is necessarily going to contain lots of sadness. If we can do something smarter (e.g. exclude the py2 block from checking on py3) then that would be better, but otherwise excluding the whole file is easier.

sigmavirus24:
So now that the tests are green, I'd like the docs to be updated appropriately to reflect all possible usages and their relevant meanings.

As for tests, we can mock out the collaborator here (urllib3) and ensure that we're passing through to it what we expect to be passed in different situations.

requests_toolbelt/adapters/source.py
```diff
@@ -28,7 +28,10 @@ class SourceAddressAdapter(HTTPAdapter):
         s.mount('http://', SourceAddressAdapter('10.10.10.10'))
     """
     def __init__(self, source_address, **kwargs):
-        self.source_address = source_address
+        if isinstance(source_address, basestring):
+            self.source_address = (source_address, 0)
+        else:
```
sigmavirus24: This should be an `elif` to check that it's a tuple.

The final else should raise a `TypeError` to prevent any further confusingly "silent" error on our part.

sigmavirus24:
I'm still very much on the fence about this, but if we are going to do this, then we need to document it to death.

requests_toolbelt/adapters/source.py
Lukasa: Do we really mandate that it's a tuple? Or is it simply necessary that it be tuple-ish? (e.g iterable of length 2?)

sigmavirus24: I'm not particular about it but I want to prevent future problems.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Lukasa:
Ok @sigmavirus24, this contains some tests and the improved docs you asked for. Thoughts?

sigmavirus24:
Looks good. One last request: Please add a test for the TypeError case.

Lukasa:
Done. =)

tests/test_source_adapter.py
```diff
@@ -0,0 +1,36 @@
...
+
+    poolmanager.PoolManager.assert_called_once_with(
+        num_pools=DEFAULT_POOLSIZE,
+        maxsize=DEFAULT_POOLSIZE,
+        block=DEFAULT_POOLBLOCK,
+        source_address=('10.10.10.10', 80)
+    )
+
+
+@patch('requests_toolbelt.adapters.source.poolmanager')
```
sigmavirus24: We're not using the patch, do we want to keep patching?


We could use it to do

```py
assert poolmanager.PoolManager.called is False
```

Lukasa: I was mostly patching to avoid the risk of touching the real thing, but yeah, let's confirm that shall we? ;)

Lukasa: Done. =)

sigmavirus24:
Thanks for putting up with me @Lukasa

Lukasa:
:heart:

PR #100, opened 2015-09-23
A few packaging niceties

untitaker:


untitaker:
You can test its behavior by running `setup.py sdist` and looking at the tarball in `dist/`.

sigmavirus24:
Thanks @untitaker

PR #106, opened 2015-10-21
Update progress bar example to use new len property

### LINKED ISSUE #105: progress_bar.py example uses old MultipartEncoder length
The progress bar example needs to be updated to use the new `len` property introduced to solve #80.

```
python progress_bar.py 
  File "/Users/dan/Downloads/progress_bar.py", line 36, in <module>
    callback = create_callback(encoder)
  File "/Users/dan/Downloads/progress_bar.py", line 15, in create_callback
    encoder_len = len(encoder)
TypeError: object of type 'MultipartEncoder' has no len()
```

DanLipsitt:
MultipartEncoder.__len__() no longer exists. Use MultipartEncoder.len instead.

Closes #105

### DIFF
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -31,3 +31,5 @@ Patches and Suggestions
 - Markus Unterwaditzer <markus@unterwaditzer.net>
 
 - Bryce Boe <bbzbryce@gmail.com> (@bboe)
+
+- Dan Lipsitt (https://github.com/DanLipsitt)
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Thanks @DanLipsitt! I guess we should be testing these too, eh?

sigmavirus24:
@Lukasa looks like the HTTPDigestAuth changes in 2.8.0 broke everything. Also our old usage of the environment variables means we still get requests 2.8.1 in our tox-env.

Lukasa:
We should probably aim to fix everything then. =( :disappointed:

PR #107, opened 2015-10-23
Attempt to install the intended version of requests with tox

sigmavirus24:


sigmavirus24:
(This is expected to fail due to problems with requests 2.8.x)

sigmavirus24:
Awesome. This now very accurately demonstrates the test failures as they should be for each of these versions. We have bugs to fix for requests 2.2.1 and 2.1.0 but those can be fixed later.

sigmavirus24:
Perfect. This is now only failing on jobs with a newer versions of requests.

PR #109, opened 2015-10-28
Fix authentication handlers using digest auth

### LINKED ISSUE #108: Classes composing or inheriting from digest auth do not function with requests 2.8.0
Requests 2.8.0 introduced a new DigestAuth class that behaves quite differently from the old one, including by having thread local storage. This invalidates the current subclasses and compositions of that class (ProxyDigestAuth, GuessAuth). They need updating.

sigmavirus24:
Requests 2.8.0 introduced a thread-safe HTTPDigestAuth class which was
backwards incompatible for anything that attempts to work with it. This
adds a compatibility layer for convenience based on the version of
requests that is installed alongside the toolbelt.

Closes #108 

---

TODO:

- [x] Fix last test
- [ ] Add more tests around this
- [x] ???

### DIFF
--- a/requests_toolbelt/auth/_digest_auth_compat.py
+++ b/requests_toolbelt/auth/_digest_auth_compat.py
@@ -1,0 +1,29 @@
+"""Provide a compatibility layer for requests.auth.HTTPDigestAuth."""
+import requests
+
+
+class _ThreadingDescriptor(object):
+    def __init__(self, prop, default):
+        self.prop = prop
+        self.default = default
+
+    def __get__(self, obj, objtype=None):
+        return getattr(obj._thread_local, self.prop, self.default)
+
+    def __set__(self, obj, value):
+        setattr(obj._thread_local, self.prop, value)
+
+
+class _HTTPDigestAuth(requests.auth.HTTPDigestAuth):
+    init = _ThreadingDescriptor('init', True)
+    last_nonce = _ThreadingDescriptor('last_nonce', '')
+    nonce_count = _ThreadingDescriptor('nonce_count', 0)
+    chal = _ThreadingDescriptor('chal', {})
+    pos = _ThreadingDescriptor('pos', None)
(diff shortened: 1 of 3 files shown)


sigmavirus24:
[This](https://travis-ci.org/sigmavirus24/requests-toolbelt/jobs/87949403#L125) is exactly what I'm seeing locally. It should get set from [this conditional](https://github.com/sigmavirus24/requests-toolbelt/blob/master/requests_toolbelt/auth/http_proxy_digest.py#L68) but appears it isn't. I'm not certain why. I think this is @Lukasa's code so memories/context would be helpful.

Lukasa:
Yup, that's what I saw. My suspicion is that we're seeing some weird threading interaction but I'm not sure what yet.

sigmavirus24:
@Lukasa so in the Proxy Digest Auth class, I had added the `init_per_thread_state()` in the `__call__` method (like it is in requests) and when I moved that to the `__init__` method, the test passed. This looks to have fixed it.

Lukasa:
Do you want to *also* add it to `__call__`? Because I think the issue here is that the auth object might be called in multiple threads *after* instantiation.

sigmavirus24:
> Do you want to also add it to `__call__`? Because I think the issue here is that the auth object might be called in multiple threads after instantiation.

So that is why I thought it belonged in `__call__`. I'll see if the tests still pass with it in both locations (but to be clear the HTTPDigestAuth class only calls that in `__call__` so I'm baffled why we would need to call it any differently.

sigmavirus24:
Eh, it works in both places and I realized that `stale_rejects` would make the proxy digest auth class not thread-safe so I took the liberty of adding extra code to protect against that by using thread local storage if it is available.

Care to give this another look over @Lukasa?

sigmavirus24:
Also, I'm not sure what tests to write for this pull request.

PR #110, opened 2015-11-13
Add ForgetfulCookieJar and Cookies module

### LINKED ISSUE #93: We should add the ForgetfulCookieJar
As discussed in kennethreitz/requests#2409, we should consider adding the occasionally-needed ForgetfulCookieJar:

```python
from requests.cookies import RequestsCookieJar

class ForgetfulCookieJar(RequestsCookieJar):
    def set_cookie(self, *args, **kwargs):
        return

s = requests.Session()
s.cookies = ForgetfulCookieJar()
```

Given that most of the function here is clearly laid out, this is an ideal contributor-friendly project. Assuming the Julython hooks are added, it's a perfect Julython task for anyone who wants it!

ceaess:
Users may want to ignore cookies during a session and can use the
ForgetfulCookieJar as a replacement for RequestsCookieJar.
Closes #93.


sigmavirus24:
:cake: :sparkles: Thank you @ceaess!

PR #112, opened 2015-11-23
Add helper for nested formdata

### LINKED ISSUE #45: form data enhancement functions
https://github.com/kennethreitz/requests/issues/2284

sigmavirus24:
Closes #45

### DIFF
--- a/requests_toolbelt/utils/formdata.py
+++ b/requests_toolbelt/utils/formdata.py
@@ -7,7 +7,7 @@ from .._compat import urlencode as _urlencode
 __all__ = ('urlencode',)
 
 
-def urlencode(query, doseq=False, safe='', encoding=None, errors=None):
+def urlencode(query, *args, **kwargs):
     """Enhanced urlencode method for handling nested queries.
 
     When using this to send data in the body of a request, make sure you
@@ -54,13 +54,7 @@ def urlencode(query, doseq=False, safe='', encoding=None, errors=None):
     while any(isinstance(v, expand_classes) for _, v in query_list):
         query_list = _expand_query_values(query_list)
 
-    try:
-        return _urlencode(query_list,
-                          doseq=doseq,
-                          encoding=encoding,
-                          errors=errors)
-    except TypeError:  # Python 2
-        return _urlencode(query_list, doseq=doseq)
+    return _urlencode(query_list, *args, **kwargs)
 
 
 def _to_kv_list(dict_or_list):
--- a/requests_toolbelt/utils/formdata.py
+++ b/requests_toolbelt/utils/formdata.py
@@ -47,7 +47,7 @@ def urlencode(query, *args, **kwargs):
     expand_classes = (dict, list, tuple)
     original_query_list = _to_kv_list(query)
 
-    if not all(len(i) for i in original_query_list):
+    if not all(len(i) == 2 for i in original_query_list):
         raise ValueError("Expected query to be able to be converted to a "
                          "list comprised of length 2 tuples.")
 
--- a/requests_toolbelt/utils/formdata.py
+++ b/requests_toolbelt/utils/formdata.py
@@ -48,7 +48,8 @@ def urlencode(query, *args, **kwargs):
     original_query_list = _to_kv_list(query)
 
     if not all(len(i) for i in original_query_list):
-        raise ValueError("Expected items to have a length of two.")
+        raise ValueError("Expected query to be able to be converted to a "
+                         "list comprised of length 2 tuples.")
 
     query_list = original_query_list
     while any(isinstance(v, expand_classes) for _, v in query_list):
--- a/requests_toolbelt/utils/formdata.py
+++ b/requests_toolbelt/utils/formdata.py
@@ -70,7 +70,7 @@ def urlencode(query, *args, **kwargs):
     expand_classes = (dict, list, tuple)
     original_query_list = _to_kv_list(query)
 
-    if not all(len(i) == 2 for i in original_query_list):
+    if not all(_is_two_tuple(i) for i in original_query_list):
         raise ValueError("Expected query to be able to be converted to a "
                          "list comprised of length 2 tuples.")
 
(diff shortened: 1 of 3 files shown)


sigmavirus24:
@Lukasa I just want to write some docs for this, but after that 0.5.0 should be shipped. I'd love some code review on this. Perhaps suggestions for warning people that this could be a rather computationally expensive operation. Also, suggestions on how to improve the expense of this.

For what it's worth I was thinking making a new list each time through the list (instead of extending) might be faster but here's some timings I ran:

```
❯❯❯ pypy -m timeit -s 'xs = [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]] * 100; y = []' -- 'for x in xs: y.extend(x)'
10000 loops, best of 3: 41.9 usec per loop
❯❯❯ pypy -m timeit -s 'xs = [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]] * 100; y = []' -- 'for x in xs: y = y + x'
100 loops, best of 3: 23.6 msec per loop

❯❯❯ python2.7 -m timeit -s 'xs = [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]] * 100; y = []' -- 'for x in xs: y.extend(x)'
100000 loops, best of 3: 14.7 usec per loop
❯❯❯ python2.7 -m timeit -s 'xs = [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]] * 100; y = []' -- 'for x in xs: y = y + x'
100 loops, best of 3: 14.2 msec per loop

❯❯❯ python3.4 -m timeit -s 'xs = [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]] * 100; y = []' -- 'for x in xs: y.extend(x)'
100000 loops, best of 3: 12.5 usec per loop
❯❯❯ python3.4 -m timeit -s 'xs = [[1, 2, 3, 4, 5, 6, 7, 8, 9, 10]] * 100; y = []' -- 'for x in xs: y = y + x'
100 loops, best of 3: 13.6 msec per loop
```

I tried to force both python 2.7 and 3.4 to do 100000 loops of the `y = y + x` timing but it was taking several minutes so I just stopped.

requests_toolbelt/utils/formdata.py
```diff
@@ -0,0 +1,81 @@
...
+    ``urllib.urlencode`` function.
+
+    .. _urlencode:
+        https://docs.python.org/3/library/urllib.parse.html#urllib.parse.urlencode
+    """
+    expand_classes = (dict, list, tuple)
+    original_query_list = _to_kv_list(query)
+
+    if not all(len(i) for i in original_query_list):
+        raise ValueError("Expected items to have a length of two.")
```
Lukasa: This ValueError is super confusing, as is this check. Mind putting a comment here or rewording the ValueError to make it clearer what's going wrong?

requests_toolbelt/utils/formdata.py
```diff
@@ -0,0 +1,81 @@
+# -*- coding: utf-8 -*-
+"""Implementation of nested form-data encoding function(s)."""
+from .._compat import basestring
+from .._compat import urlencode as _urlencode
+
+
+__all__ = ('urlencode',)
+
+
+def urlencode(query, doseq=False, safe='', encoding=None, errors=None):
```
Lukasa: Given that you don't touch the other arguments directly, is there any reason this shouldn't be `def urlencode(query, *args, **kwargs)`? This will allow you to avoid the `try..except` block further down the page.

requests_toolbelt/utils/formdata.py
```diff
@@ -0,0 +1,81 @@
...
+
+def _expand_query_values(original_query_list):
+    query_list = []
+    for key, value in original_query_list:
+        if isinstance(value, basestring):
+            query_list.append((key, value))
+        else:
+            key_fmt = key + '[%s]'
+            value_list = _to_kv_list(value)
+            query_list.extend([(key_fmt % k, v) for k, v in value_list])
```
Lukasa: Do you need to create an intermediate list here? You can remove the list comprehension, have this as a generator comprehension instead, and avoid the intermediate list that needs to be allocated and then immediately deallocated.

tests/test_formdata.py
```diff
@@ -0,0 +1,36 @@
...
+            'third_nested': {
+                'fourth0': 'fourth_value0',
+                'fourth1': 'fourth_value1',
+            },
+            'third0': 'third_value0',
+        },
+        'second0': 'second_value0',
+    },
+    'outter': 'outter_value',
+}
```
Lukasa: This test doesn't actually include a list or a tuple, and I'm willing to bet that if/when you add those types you won't get the end result you expect. =)

Lukasa:
This is generally good, aside from the fact that I don't think it handles lists/tuples properly. Otherwise, a few notes. =)

requests_toolbelt/utils/formdata.py
sigmavirus24: `*args` and `**kwargs` don't lend themselves to discoverability but I understand your point. I may as well just do that.

requests_toolbelt/utils/formdata.py
sigmavirus24: Yep. This is sleepy Ian writing code error messages. Thanks for catching that.

requests_toolbelt/utils/formdata.py
sigmavirus24: Fixed.

tests/test_formdata.py
sigmavirus24: Oh really? See the results of eb5e154 and f1d9b82. ;P

requests_toolbelt/utils/formdata.py
sigmavirus24: Done

requests_toolbelt/utils/formdata.py
sigmavirus24: Let me know if the new message is better. I jsut realized why the check is confusing. Coffee is finally kicking in.

Lukasa:
\o/ LGTM.

PR #114, opened 2015-11-24
Add AppEngineAdapter for GAE users

sigmavirus24:
At the moment this will not work because the AppEngineManager does not
have the same signature as urllib3's PoolManager.

Related to:
https://github.com/kennethreitz/requests/issues/1905#issuecomment-159380704

cc @Lukasa @shazow @jonparrott @webmaven @jacobg

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,40 @@
 History
 =======
 
+0.6.0 -- 2016-xx-yy
+-------------------
+
+More information about this release can be found on the `0.6.0 milestone`_.
+
+New Features
+~~~~~~~~~~~~
+
+- Add ``AppEngineAdapter`` to support developers using Google's AppEngine
+  platform with Requests.
+
+- Add ``GuessProxyAuth`` class to support guessing between Basic and Digest
+  Authentication for proxies.
+
+Fixed Bugs
+~~~~~~~~~~
+
+- Ensure that proxies use the correct TLS version when using the
+  ``SSLAdapter``.
+
+- Fix an ``AttributeError`` when using the ``HTTPProxyDigestAuth`` class.
+
+Miscellaneous
+~~~~~~~~~~~~~
+
+- Drop testing support for Python 3.2. virtualenv and pip have stopped
+  supporting it meaning that it is harder to test for this with our CI
+  infrastructure. Moving forward we will make a best-effort attempt to
+  support 3.2 but will not test for it.
+
+
+.. _0.6.0 milestone:
+    https://github.com/sigmavirus24/requests-toolbelt/milestones/0.6.0
+
 0.5.1 -- 2015-12-16
 -------------------
 
--- a/docs/adapters.rst
+++ b/docs/adapters.rst
@@ -7,6 +7,8 @@ The toolbelt comes with several different transport adapters for you to use
 with requests. The transport adapters are all kept in
 :mod:`requests_toolbelt.adapters` and include
 
+- :class:`requests_toolbelt.adapters.appengine.AppEngineAdapter`
+
 - :class:`requests_toolbelt.adapters.fingerprint.FingerprintAdapter`
 
 - :class:`requests_toolbelt.adapters.socket_options.SocketOptionsAdapter`
@@ -16,6 +18,40 @@ with requests. The transport adapters are all kept in
 - :class:`requests_toolbelt.adapters.source.SourceAddressAdapter`
 
 - :class:`requests_toolbelt.adapters.ssl.SSLAdapter`
+
+AppEngineAdapter
+----------------
+
+.. versionadded:: 0.6.0
+
+As of version 2.10.0, Requests will be capable of supporting Google's App
+Engine platform. In order to use Requests on GAE, however, you will need a
+custome adapter found here as
+:class:`~requests_toolbelt.adapters.appengine.AppEngineAdapter`. There are two
+ways to take advantage of this support at the moment:
+
+#. Using the :class:`~requests_toolbelt.adapters.appengine.AppEngineAdapter`
+   like every other adapter, e.g.,
+
+   .. code-block:: python
+
+       import requests
+       from requests_toolbelt.adapters import appengine
+
+       s = requests.Session()
+       s.mount('http://', appengine.AppEngineAdapter())
+       s.mount('https://', appengine.AppEngineAdapter())
+
+#. By monkey-patching requests to always use the provided adapter:
+
+   .. code-block:: python
+
+       import requests
+       from requests_toolbelt.adapters import appengine
+
+       appengine.monkeypatch()
+
+.. autoclass:: requests_toolbelt.adapters.appengine.AppEngineAdapter
 
 FingerprintAdapter
 ------------------
--- a/requests_toolbelt/_compat.py
+++ b/requests_toolbelt/_compat.py
@@ -11,6 +11,8 @@ urllib3 without providing a shim.
 from collections import Mapping, MutableMapping
 import sys
 
+import requests
+
 try:
     from requests.packages.urllib3 import connection
     from requests.packages.urllib3 import fields
@@ -21,6 +23,14 @@ except ImportError:
     from urllib3 import fields
     from urllib3 import filepost
     from urllib3 import poolmanager
+
+if requests.__build__ < 0x020800:
+    gaecontrib = None
+else:
+    try:
+        from requests.packages.urllib3.contrib import appengine as gaecontrib
+    except ImportError:
+        from urllib3.contrib import appengine as gaecontrib
 
 PY3 = sys.version_info > (3, 0)
 
@@ -275,4 +285,5 @@ __all__ = (
     'HTTPHeaderDict',
     'queue',
     'urlencode',
+    'gaecontrib',
 )
--- a/requests_toolbelt/adapters/appengine.py
+++ b/requests_toolbelt/adapters/appengine.py
@@ -32,7 +32,7 @@ class AppEngineAdapter(adapters.HTTPAdapter):
         >>>
     """
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, validate_certificate=True, *args, **kwargs):
         if gaecontrib is None:
             raise exc.VersionMismatchError(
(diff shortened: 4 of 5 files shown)


jonparrott:
What's the issue with the signature?

sigmavirus24:
[PoolManager.__init__](https://github.com/shazow/urllib3/blob/3bfcd94a3f70fed82cd2a20f8a065816ea092e55/urllib3/poolmanager.py#L62) and [AppEngineManager.__init__](https://github.com/shazow/urllib3/blob/3bfcd94a3f70fed82cd2a20f8a065816ea092e55/urllib3/contrib/appengine.py#L59) take different parameters. It'd be convenient if `AppEngineManager` at least accepted the same parameters (even if as a `*args, **kwargs` set of parameters that aren't used).

shazow:
I think it makes sense for the constructors to be relative to the backend they're implementing, as long as the instance's interface is sufficiently common. I'd be a strong -1 against accepting arguments that are not relevant/not implemented and simply ignoring them.

If anything, could add a "backport"-ish constructor helper, but honestly I think that's the job of the adapter.

jonparrott:
+1 to @shazow.

Lukasa:
The bigger problem here is that this won't work because of the way Requests uses the poolmanager. We assume it has the methods `connection_from_url()` and `clear()`, and then also expect that `connection_from_url` returns something that has the following properties and methods:

- `cert_reqs`
- `ca_certs`
- `ca_cert_dir`
- `urlopen`
- `_get_conn`
- `putrequest`
- `putheader`
- `endheaders`
- `send`
- `getresponse`

The `AppEngineManager` simply does not work that way. To get this to work, we need to rewrite `send` as well. =)

jonparrott:
`connection_from_url` and `clear` are easy to mock out in an adapter.

We already implement `urlopen`, but I'm unsure how to go about the others at this time.

sigmavirus24:
Yeah this also isn't the highest on my priority list at the moment.

jonparrott:
@Lukasa do we have any tests that show that requests requires all of those attributes/methods?

@sigmavirus24 I can probably take a poke at it over the long weekend, but urllib/requests internals are not my speciality (but App Engine is!).

Lukasa:
I think the better approach is to acknowledge that AppEngine is unlike other cases and just override `send`. That removes the need for the other stuff because those are only called from `send`.

Sadly @jonparrott we don't, and in fact testing on GAE may be tricky because we run an in-process web server to test against.

shazow:
Yea, I think the problem is that Requests tries to serialize/encode the request itself, then bypass much of urllib3 and just use the pooling mechanism to send the request. The whole point of the AppEngineManager is that there is no pooling—it uses AppEngine's fetchurl mechanism which abstracts all the pooling and all that.

I'm not familiar enough with Requests' internal API but I have a feeling it might be less work to make a custom AppEngine driver for Requests that bypasses much of Requests' logic rather than trying to mock out urllib3's AppEngineManager's lack of lower-level stuff.

jonparrott:
> Sadly @jonparrott we don't, and in fact testing on GAE may be tricky because we run an in-process web server to test against.

@Lukasa sorry I was unclear, I wasn't looking for GAE tests, I was looking for any other tests that demonstrates this - something to build off of.

Lukasa:
@jonparrott Yeah, I know. =) What I was saying is: even if you could find the appropriate tests, they'd inevitably be using an in-process web server that would make testing in a GAE-like environment really tricky.

The reality is that GAE is just not a priority for the requests project at all.

jonparrott:
> The reality is that GAE is just not a priority for the requests project at all.

I'm well aware and they are justified in such a decision. However, if there's anything I can personally do to make it possible to use requests on GAE, I can put that in my priorities. Ensuring that AppEngineManager in urllib3 continues to work is part of my priorities already.

> make testing in a GAE-like environment really tricky.

I've managed to get a subset of tests working for urllib3 on within GAE, so hopefully I can manage some for requests. I'll see when I can take a look at this, as there's a lot of internals I need to learn to make this work.

mikelambert:
I've submitted a pull request for https://github.com/shazow/urllib3/pull/763, which should fix the issues preventing AppEngineManager from working properly, and addresses many of the issues in this thread (though I do accept dummy params for calling consistency).

Ideally when combined with https://github.com/sigmavirus24/requests-toolbelt/pull/119 to feature/gae, I would then also push for this feature/gae being merged into master branch.

As far as tests, I have not done any work to set up an appengine test case (beyond my use a Google library GCM that depends on the requests library, which was previously failing when running on AppEngine URLFetch). However, it should be possible to build a skeleton standalone AppEngine app that exercises basic requests functionality, which can be run with the AppEngine dev appserver (or even within nosetests!) to verify things continue working.

Though honestly I'd prefer @jonparrott build this test app if possible, simply so Google has more ownership/knowledge of GAE's support of the requests library. I quit Google earlier this year to work on my own stuff, so feels ironic that I'm now doing work that I (or Jon) could have gotten paid to do while at Google. ;)

jonparrott:
Thanks for picking up the mantle here, @mikelambert. 

I personally think that we should add the compatibility shims here, keep the chunked encoding fix in [shazow/urllib3#763](https://github.com/shazow/urllib3/pull/763), and then do #119.

> Though honestly I'd prefer @jonparrott build this test app if possible

I have the basic harness for GAE + urllib3, I can push that up. @shazow @sigmavirus24 where would be the best place for this to live? It's essentially a standalone GAE app. If we're going to expand it to include requests tests, should it live here?

> so feels ironic that I'm now doing work that I (or Jon) could have gotten paid to do while at Google. ;)

Luckily, I am paid to work on this (my assignment is "Python on Google Cloud Platform"). Unfortunately Q4 has been busy for me, but I can start giving this much more attention in January.

mikelambert:
I'm a bit confused, exactly what compatibility shims are you proposing I move/add into requests_toolbelt's appengine.py? If the shim is making AppEngineManager behave more like urllib3's PoolManager (in terms of supporting clear() and connection_from_url()), then why *wouldn't* we want that to live in urllib3?

As far as the basic test harness, I don't really understand the "requests includes urllib3 and there's this external requests_toolbelt library too" setup, so no votes from me on where it should belong. Though if it's just an app that runs tests, that sounds great. (Which if combined with nosetests --with-gae that starts a dev appserver to run tests against it, sounds even better for inclusion into an existing test suite.)

jonparrott:
> then why wouldn't we want that to live in urllib3?

Because afaik, urllib3 does not define any interface for Pools except for `urlopen`, which AppEnginePool satisfies. It's requests that assumes that all managers accept the same arguments and have the same methods (erroneously). Therefore, it leads that we shouldn't "pollute" the AppEngineManager with compatibility code for requests. That's what toolbelt is for.

> dev appserver to run tests against it, sounds even better for inclusion into an existing test suite.

dev_appserver != real gae, especially in this case. The tests in urllib3 for AppEngineManager already cover this. I have an actual deployable app that runs a subset of tests in production.

mikelambert:
But my argument is that AppEnginePool does not satisfy that API. In particular, any unrecognized parameter (such as `assert_same_host`, which is part of this API I believe?) gets slurped into `\*\*response_kw`, and passed to `HTTPResponse(..., \*\*response_kw)`, which then breaks.

mikelambert:
Oh, and while dev_appserver != real gae, all of these fixes were produced with dev_appserver testing. I use Managed VMs in prod, so can just use sockets directly and avoid this whole mess, but my development system is busted without these fixes.

So a test suite that runs in nosetests with dev_appserver, while not as representative and accurate as prod, may prove to help keep compatibility working in the face of further development, simply due to the automated nature of it. I'm basically arguing for both dev and prod tests: ie, a shim around your prod tests that get them working in dev appserver with nosetests, so that automated test frameworks can ensure things don't break too horribly.

jonparrott:
> arguing for both dev and prod tests

we're on the same page here.

jacobg:
I'm by no means very knowledgeable about any of this stuff, but I did observe in a recent version of Requests that the use of sockets called in production App Engine works to non-Google endpoints, whereas Google endpoints fails. This is because their beta sockets features blocks sockets connections to Google endpoints. It's all explained here:   https://cloud.google.com/appengine/docs/python/sockets/
See "Limitations and restrictions" section.

mikelambert:
@jacobg Yup, very good point, and I believe that's true for standard prod AppEngine if you decide to use raw sockets. However, I am using Managed VMs, which are basically just containers running on the GCE infrastructure, and I *believe* they are exempt from the "No Google endpoints" rule. I haven't heard any official direction on this, but I am able to send Google Cloud Messaging notifications to my Android device from my GAE Managed VM app that's using raw sockets.

jonparrott:
> @jacobg: Requests that the use of sockets in production App Engine works to non-Google endpoints, whereas Google endpoints fail.

This is a known limitation documented in GAE docs. I can also add it to the urllib3 docs [here](https://urllib3.readthedocs.org/en/latest/contrib.html#google-app-engine), if you'd like. It would also make sense for us to mention that in the documentation for requests-toolbelt.

> @mikelambert: I believe they are exempt from the "No Google endpoints" rule. I haven't heard any official direction on this.

Managed VMs uses compute engine networking, and there should be no issues accessing Google APIs from GCE. I'm sorry if this is unclear in the documentation, I'll try to fix that.

@mikelambert If you're using Managed VMs, why do you need any compatibility shim for requests? You should be able to use the ordinary, unmodified library.

mikelambert:
I want to be able to test my code in the dev_appserver, which AFAICT doesn't support raw sockets. Thus my involvement and write-up of https://github.com/sigmavirus24/requests-toolbelt/pull/119 (feel free to chime in there, as it's not yet committed...)

jonparrott:
> I want to be able to test my code in the dev_appserver

Gotcha, so you are still using `google.appengine.*` APIs.

sigmavirus24:
Note: This still needs documentation to be written.

mikelambert:
Is README.rst the correct place for this documentation? Does anything else out there need fixing up, in requests' README or elsewhere?

sigmavirus24:
@mikelambert I already added documentation.

PR #118, opened 2015-12-16
Implement our own total_len function

### LINKED ISSUE #117: MultipartEncoder needs to implement it's own total_length function
Currently the encoder uses requests' `super_len` function which changed in 2.9.0. This broke uploads using the encoder, especially in tools like twine: https://github.com/kennethreitz/requests/issues/2929#issuecomment-165111085

sigmavirus24:
requests 2.9.0 changed super_len and broke the MultipartEncoder. To
prevent this, we need to implement our own function to get the total
length of an object.

Closes #117

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,21 @@
 History
 =======
 
+0.5.1 -- 2015-12-16
+-------------------
+
+More information about this release can be found on the `0.5.1 milestone`_.
+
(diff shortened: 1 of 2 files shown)


sigmavirus24:
@Lukasa would you mind giving this a once over?

Lukasa:
LGTM! :sparkles: :cake: :sparkles:

PR #119, opened 2015-12-17
Adds a monkeypatch() function that ensures all Sessions use the AppEngineAdapter compatibility class.

mikelambert:
The goal here is to allow the following code to work:
```
from requests_toolbelt.adapters import appengine
appengine.monkeypatch()
```

...and then run all your libraries that have dependencies on requests and sessions buried deep inside, all working properly with URLFetch on Appengine.

The code doesn't work as-is, as it depends on the following change:
https://github.com/shazow/urllib3/pull/763
...which then would probably need to be released and ported into the requests library as well...not sure how long that takes or what the process is.

This also is tangentially related to the following issues relating to getting requests playing nicely with AppEngine's URLFetch:
https://github.com/kennethreitz/requests/issues/1905
https://github.com/shazow/urllib3/issues/618

### DIFF
--- a/requests_toolbelt/_compat.py
+++ b/requests_toolbelt/_compat.py
@@ -24,7 +24,15 @@ except ImportError:
     from urllib3 import filepost
     from urllib3 import poolmanager
 
-if requests.__build__ < 0x020800:
+if requests.__build__ < 0x020300:
+    timeout = None
+else:
+    try:
+        from requests.packages.urllib3.util import timeout
+    except ImportError:
+        from urllib3.util import timeout
+
+if requests.__build__ < 0x021000:
     gaecontrib = None
 else:
     try:
@@ -282,6 +290,7 @@ __all__ = (
     'fields',
     'filepost',
     'poolmanager',
+    'timeout',
     'HTTPHeaderDict',
     'queue',
     'urlencode',
--- a/requests_toolbelt/adapters/appengine.py
+++ b/requests_toolbelt/adapters/appengine.py
@@ -1,49 +1,139 @@
 # -*- coding: utf-8 -*-
 """The App Engine Transport Adapter for requests.
 
-This requires a version of requests >= 2.8.0.
+This requires a version of requests >= 2.10.0.
 """
 import requests
 from requests import adapters
+from requests import sessions
 
 from .. import exceptions as exc
 from .._compat import gaecontrib
+from .._compat import timeout
+
+"""
+.. versionadded:: 0.6.0
+
+There are two ways to use this library.
+
+If you're using requests directly, you can use code like:
+    >>> import requests
+    >>> import ssl
+    >>> from requests.packages.urllib3.contrib import appengine as ul_appengine
+    >>> from requests_toolbelt.adapters import appengine
+    >>> s = requests.Session()
+    >>> if ul_appengine.is_appengine_sandbox():
+    ...    s.mount('http://', appengine.AppEngineAdapter())
+    ...    s.mount('https://', appengine.AppEngineAdapter())
+
+If you depend on external libraries which use requests, you can use code like:
+    >>> from requests_toolbelt.adapters import appengine
+    >>> appengine.monkeypatch()
+which will ensure all requests.Session objects use AppEngineAdapter properly.
+"""
 
 
 class AppEngineAdapter(adapters.HTTPAdapter):
     """A transport adapter for Requests to use urllib3's GAE support.
 
+    Implements request's HTTPAdapter API.
+
     When deploying to Google's App Engine service, some of Requests'
     functionality is broken. There is underlying support for GAE in urllib3.
     This functionality, however, is opt-in and needs to be enabled explicitly
     for Requests to be able to use it.
-
-    Example usage:
-
-    .. code-block:: python
-
-        >>> import requests
-        >>> import ssl
-        >>> from requests_toolbelt.adapters import appengine
-        >>> s = requests.Session()
-        >>> if using_appengine():
-        ...    s.mount('https://', appengine.AppEngineAdapter())
-        ...
-        >>>
     """
 
     def __init__(self, validate_certificate=True, *args, **kwargs):
-        if gaecontrib is None:
-            raise exc.VersionMismatchError(
-                "The toolbelt requires at least Requests 2.8.0 to be "
-                "installed. Version {0} was found instead.".format(
-                    requests.__version__
-                )
-            )
+        _check_version()
         self._validate_certificate = validate_certificate
         super(AppEngineAdapter, self).__init__(self, *args, **kwargs)
 
     def init_poolmanager(self, connections, maxsize, block=False):
-        self.poolmanager = gaecontrib.AppEngineManager(
-            validate_certificate=self._validate_certificate
+        self.poolmanager = _AppEnginePoolManager(self._validate_certificate)
+
+
+class _AppEnginePoolManager(object):
+    """Implements urllib3's PoolManager API expected by requests.
+
+    While a real PoolManager map hostnames to reusable Connections,
+    AppEngine has no concept of a reusable connection to a host.
+    So instead, this class constructs a small Connection per request,
+    that is returned to the Adapter and used to access the URL.
+    """
+
+    def __init__(self, validate_certificate=True):
+        self.appengine_manager = gaecontrib.AppEngineManager(
+            validate_certificate=validate_certificate)
+
+    def connection_from_url(self, url):
+        return _AppEngineConnection(self.appengine_manager, url)
+
+    def clear(self):
+        pass
+
+
+class _AppEngineConnection(object):
+    """Implements urllib3's HTTPConnectionPool API's urlopen().
+
+    This Connection's urlopen() is called with a host-relative path,
+    so in order to properly support opening the URL, we need to store
+    the full URL when this Connection is constructed from the PoolManager.
+
+    This code wraps AppEngineManager.urlopen(), which exposes a different
+    API than in the original urllib3 urlopen(), and thus needs this adapter.
+    """
+
+    def __init__(self, appengine_manager, url):
+        self.appengine_manager = appengine_manager
+        self.url = url
+
+    def urlopen(self, method, url, body=None, headers=None, retries=None,
+                redirect=True, assert_same_host=True,
+                timeout=timeout.Timeout.DEFAULT_TIMEOUT,
+                pool_timeout=None, release_conn=None, **response_kw):
+        # This function's url argument is a host-relative URL,
+        # but the AppEngineManager expects an absolute URL.
+        # So we saved out the self.url when the AppEngineConnection
+        # was constructed, which we then can use down below instead.
+        # Let's verify our assumptions here though, just in case.
+        assert self.url.endswith(url), (
+            "AppEngineConnection constructed "
+            "with (%s), and called urlopen with (%s). Expected the latter "
+            "to be the host-relative equivalent of the former." %
+            (self.url, url))
+
+        # Jump through the hoops necessary to call AppEngineManager's API.
+        return self.appengine_manager.urlopen(
+            method,
+            self.url,
+            body=body,
+            headers=headers,
+            retries=retries,
+            redirect=redirect,
+            timeout=timeout,
+            **response_kw)
+
+
+def monkeypatch():
+    """Sets up all Sessions to use AppEngineAdapter by default.
+
+    If you don't want to deal with configuring your own Sessions,
+    or if you use libraries that use requests directly (ie requests.post),
+    then you may prefer to monkeypatch and auto-configure all Sessions.
+    """
+    _check_version()
+    # HACK: We should consider modifying urllib3 to support this cleanly,
+    # so that we can set a module-level variable in the sessions module,
+    # instead of overriding an imported HTTPAdapter as is done here.
+    sessions.HTTPAdapter = AppEngineAdapter
+
+
+def _check_version():
+    if gaecontrib is None:
+        raise exc.VersionMismatchError(
+            "The toolbelt requires at least Requests 2.10.0 to be "
+            "installed. Version {0} was found instead.".format(
+                requests.__version__
+            )
         )


mikelambert:
And once/if this is merged, I'll request that feature/gae be merged into master as well.

Lukasa:
I think this is a reasonable stop-gap solution to this problem. It would be nice if there was a better way to do this though. Class-level variables on the `Session`, maybe?

mikelambert:
I agree class-level variables on the Session would be better. However, even if we fixed that in requests library, we'd still probably want need to support backwards-compatibility with older versions of the request library, so this code would still be necessary anyway, right? So I viewed that improvement as a non-blocker for this change.

Lukasa:
Yup, it's definitely a non-blocker.

mikelambert:
As requested in https://github.com/shazow/urllib3/pull/763 , I've moved most of that complexity into this requests-toolbelt patch instead. The urllib3 changes have been merged, so hopefully @sigmavirus24 or @Lukasa can have another look here. Thanks!

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +49,57 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
+
+    def __init__(self, pool_manager, url):
+        self.pool_manager = pool_manager
+        self.url = url
+
+    def urlopen(self, method, url, body=None, headers=None, retries=None,
+                redirect=True, assert_same_host=True, timeout=timeout.Timeout.DEFAULT_TIMEOUT,
+                pool_timeout=None, release_conn=None, **response_kw):
+        # Jump through the hoops necessary to call AppEngineManager's API.
+        appengine_manager = gaecontrib.AppEngineManager(
```
jonparrott: wouldn't it be better to just create this once?

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +49,57 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
+
+
+class _AEPoolManager(object):
+    """Implements urllib3's PoolManager API."""
+
+    def __init__(self, validate_certificate=True):
+        self.validate_certificate = validate_certificate
+
+    def connection_from_url(self, url):
+        return _AEConnection(self, url)
```
jonparrott: Couldn't you create only one connection and return it every time instead of allocating new connections?

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +49,57 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
         super(AppEngineAdapter, self).__init__(self, *args, **kwargs)
 
     def init_poolmanager(self, connections, maxsize, block=False):
-        self.poolmanager = gaecontrib.AppEngineManager(
-            validate_certificate=self._validate_certificate
-        )
+        self.poolmanager = _AEPoolManager(self._validate_certificate)
+
+
+class _AEPoolManager(object):
```
jonparrott: Nit: Why not `_AppEnginePoolManager`? I'm not a fan of abbreviations like this. (but this is also not my project. ;))

requests_toolbelt/adapters/appengine.py
```diff
@@ -27,6 +31,7 @@ class AppEngineAdapter(adapters.HTTPAdapter):
         >>> from requests_toolbelt.adapters import appengine
         >>> s = requests.Session()
         >>> if using_appengine():
```
jonparrott: Where does the `using_appengine` function come from?

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +49,57 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
 
     def init_poolmanager(self, connections, maxsize, block=False):
-        self.poolmanager = gaecontrib.AppEngineManager(
-            validate_certificate=self._validate_certificate
-        )
+        self.poolmanager = _AEPoolManager(self._validate_certificate)
+
+
+class _AEPoolManager(object):
+    """Implements urllib3's PoolManager API."""
```
jonparrott: Can you explain in this docstring why this is necessary? We know due to the discussions, but it's nice to have it documented along with the code.

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +49,57 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
+
+    def urlopen(self, method, url, body=None, headers=None, retries=None,
+                redirect=True, assert_same_host=True, timeout=timeout.Timeout.DEFAULT_TIMEOUT,
+                pool_timeout=None, release_conn=None, **response_kw):
+        # Jump through the hoops necessary to call AppEngineManager's API.
+        appengine_manager = gaecontrib.AppEngineManager(
+            validate_certificate=self.pool_manager.validate_certificate)
+        return appengine_manager.urlopen(
+            method,
+            self.url, # We need to use the full URL here, not the host-relative path
```
jonparrott: Can we have an assertion that `url` and `self.url` are 'equivalent'?

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +49,57 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
+    """
+
+    def __init__(self, pool_manager, url):
+        self.pool_manager = pool_manager
+        self.url = url
+
+    def urlopen(self, method, url, body=None, headers=None, retries=None,
+                redirect=True, assert_same_host=True, timeout=timeout.Timeout.DEFAULT_TIMEOUT,
+                pool_timeout=None, release_conn=None, **response_kw):
+        # Jump through the hoops necessary to call AppEngineManager's API.
```
jonparrott: Can we document why this is necessary? See earlier comment along the same lines.

seawolf42:
I'm confused and hoping someone can clarify this for me. I've checked out requests-toolbelt, applied patches 114 and 119, and am referencing that modified version of requests-toolbelt from my project. I am unclear whether I am supposed to also apply 763 to urllib3, I have not yet done this.

I also added the `appengine.monkeypatch()` call in main.py prior to importing or calling anything else within my application.

When I try to run my project, I get:

    'Timeout' object has no attribute 'read'
    ... most of the stack trace removed ...
    .../requests/packages/urllib3/contrib/appengine.py in _get_absolute_timeout
    180.        if timeout.read is not timeout.connect: ...

I feel like I'm missing something obvious here... can someone point me at my oversight?

mikelambert:
@seawolf42 Yes, you need to patch 763 into request's version of urllib3.

For background, 763 was merged into urllib3, but requests has its own version of urllib3, and the urllib3 head hasn't been integrated there yet. Once the requests-toolbelt changes go in, I will push harder on getting someone to merge that.

requests_toolbelt/adapters/appengine.py
mikelambert: No, the connection needs to be specific to the url itself:

This url contains the full path including hostname. The AEConnection's urlopen() is called later, but it is called with a host-relative URL that is missing the hostname. The only way to get the hostname is from the connection, which was initialized with the url-including-hostname. So that's my argument as to why I cannot reuse these...unless I'm missing something?

requests_toolbelt/adapters/appengine.py
mikelambert: Not sure, this docstring predates me in @sigmavirus24 's original pull request. I'm personally not a fan of this approach anyway, preferring the monkeypatch() approach below, so that my third-party libraries I use automatically get working Sessions.

requests_toolbelt/adapters/appengine.py
mikelambert: Hmmm...I could construct an actual pool of Connections inside the PoolManager, indexed by hostname, and then return the right connection based on the incoming url's host (constructing one if necessary.) And then inside the connection's urlopen, concatenate the connection's hostname with the incoming host-relative path.

Do you think that code complexity is worth it, for the gain of reusing the AppEngineConnection and AppEngineManager objects on multiple requests to the same host? (There's no actual "connection" that's getting re-used though, since it's calling urlfetch repeatedly under the hood either way.)

requests_toolbelt/adapters/appengine.py
mikelambert: Good idea, fixed. Though really, the way this code gets used, I believe urlopen was only called once per connection anyway. (Though see note in next comment, as that assumption could change.)

requests_toolbelt/adapters/appengine.py
mikelambert: Fixed.

requests_toolbelt/adapters/appengine.py
mikelambert: I'm unsure what you want me to document here...(or down below, either)

Before, init_poolmanager was creating an AppEngineManager, which was not actually a pool manager, and didn't actually work at all. Now init_poolmanager is creating an AppEnginePoolManager, and the poolmanager's connection_from_url constructs an AppEngineConnection, which seems to be straightforward and obvious. The classes are really only needed to implement the APIs that requests library expects of the AppEngine adapters, and the docstrings already say that...

Maybe I'm too deep in this to see what needs documenting here, but if you give me an example of what you were thinking docstring-wise, I can fill it out...

requests_toolbelt/adapters/appengine.py
mikelambert: I expanded the comment, and added an assertion that captures the invariant of "equivalent" that's expected here.

requests_toolbelt/adapters/appengine.py
sigmavirus24: It's not meant to be literal code. It's pseudo-code and many people abhor monkey-patching libraries @mikelambert so it is an example for them as to how to do this.

jonparrott: Why not use the real function [`is_appengine_sandbox`](https://github.com/shazow/urllib3/blob/master/urllib3/contrib/appengine.py#L215)? IMO, it's better to have an example that works.

(Don't let this block though, I can always submit a PR to update docstrings)

requests_toolbelt/adapters/appengine.py
jonparrott: Just something along the lines of "Implements the PoolManager API expected by requests. The differences are...".

requests_toolbelt/adapters/appengine.py
jonparrott: Not sure, it depends on how requests treats connection objects. App Engine has a relatively low memory limit (128mb) so it's better to avoid unnecessary allocations, but it's also not good to accumulate a ton of objects in a indeterminate pool.

I think creating a connection per requests is fine, but please re-use the underlying manager.

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +49,66 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
+class _AppEngineConnection(object):
+    """Implements urllib3's HTTPConnectionPool API's urlopen().
+
+    This Connection's urlopen() is called with a host-relative path,
+    so in order to properly support opening the URL, we need to store the full URL
+    when we are constructed from the Manager.
+    """
+
+    def __init__(self, pool_manager, url):
+        self.appengine_manager = gaecontrib.AppEngineManager(
```
jonparrott: Could you have the `pool_manager` construct this and pass it down to each connection?

requests_toolbelt/adapters/appengine.py
mikelambert: Hopefully this should be much better now.

requests_toolbelt/adapters/appengine.py
mikelambert: Great idea, fixed.

requests_toolbelt/adapters/appengine.py
mikelambert: I've reworked the comment to use is_appengine_sandbox(), which in addition to working code, also handles Managed VMs properly.

jonparrott: :+1:

jonparrott:
@Lukasa, @sigmavirus24 I'm satisfied with this, if you guys want to take a look.

seawolf42:
@mikelambert, thanks, that did the trick! I can confirm that using just these three patches (114 and 119 on requests-toolbelt and 763 on urllib3 within request) I am able to use python-social-auth on GAE with DjangoAppEngine, at least within the dev server.

Lukasa:
This looks reasonable to me. =)

jonathan-s:
@seawolf42 Did you succeed in production as well not only on the dev server?

sigmavirus24:
@jonathan-s have you tested this and found problems? If so, please let us know so we can have this in the best shape possible before merging.

jonathan-s:
I haven't tested it, I'll try to make some tests and see what I get.

jonathan-s:
@sigmavirus24 When I tried it in devapp_server with the following code
    
    from requests_toolbelt.adapters import appengine
    from requests_toolbelt.adapters.appengine import AppEngineAdapter
    appengine.monkeypatch()

    import requests
    import webapp2
    from settings import JINJA_ENVIRONMENT

    class RequestsView(webapp2.RequestHandler):

        def get(self):
            url = 'http://www.example.com'
            session = requests.Session()
            # response = session.mount(url, AppEngineAdapter())
            response = requests.get(url)

            tmp_values = {}
            tmp_values['status'] = response.status_code
            tmp_values['content'] = response.content[:500]

            template = JINJA_ENVIRONMENT.get_template('templates/requests_view.html')
            self.response.write(template.render(tmp_values))

I got this traceback: 

      File "/Users/jonathan/projects/honshuu-console/lib/webapp2.py", line 1535, in __call__
        rv = self.handle_exception(request, response, e)
      File "/Users/jonathan/projects/honshuu-console/lib/webapp2.py", line 1529, in __call__
        rv = self.router.dispatch(request, response)
      File "/Users/jonathan/projects/honshuu-console/lib/webapp2.py", line 1278, in default_dispatcher
        return route.handler_adapter(request, response)
      File "/Users/jonathan/projects/honshuu-console/lib/webapp2.py", line 1102, in __call__
        return handler.dispatch()
      File "/Users/jonathan/projects/honshuu-console/lib/webapp2.py", line 572, in dispatch
        return self.handle_exception(e, self.app.debug)
      File "/Users/jonathan/projects/honshuu-console/lib/webapp2.py", line 570, in dispatch
        return method(*args, **kwargs)
      File "/Users/jonathan/projects/honshuu-console/views/requests_view.py", line 18, in get
        response = requests.get(url)
      File "/Users/jonathan/projects/honshuu-console/lib/requests/api.py", line 67, in get
        return request('get', url, params=params, **kwargs)
      File "/Users/jonathan/projects/honshuu-console/lib/requests/api.py", line 53, in request
        return session.request(method=method, url=url, **kwargs)
      File "/Users/jonathan/projects/honshuu-console/lib/requests/sessions.py", line 468, in request
        resp = self.send(prep, **send_kwargs)
      File "/Users/jonathan/projects/honshuu-console/lib/requests/sessions.py", line 576, in send
        r = adapter.send(request, **kwargs)
      File "/Users/jonathan/projects/honshuu-console/lib/requests/adapters.py", line 376, in send
        timeout=timeout
      File "/Users/jonathan/projects/honshuu-console/lib/requests_toolbelt/adapters/appengine.py", line 116, in urlopen
        **response_kw)
      File "/Users/jonathan/projects/honshuu-console/lib/requests/packages/urllib3/contrib/appengine.py", line 105, in urlopen
        deadline=self._get_absolute_timeout(timeout),
      File "/Users/jonathan/projects/honshuu-console/lib/requests/packages/urllib3/contrib/appengine.py", line 180, in _get_absolute_timeout
        if timeout.read is not timeout.connect:
    AttributeError: 'Timeout' object has no attribute 'read'

jonathan-s:
Seeing now that it might be because of the internal use of request which uses an earlier version urllib3

seawolf42:
This is working great in production with both python-social-auth and facebook-sdk packages.

jonathan-s:
I got it working as well.

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +53,75 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
+
+
+def monkeypatch():
+    """Sets up all Sessions to use AppEngineAdapter by default.
+
+    If you don't want to deal with constructing and configuring your own Sessions,
+    or if you use libraries that call the requests API directly (ie requests.post),
+    then you may prefer to monkeypatch and auto-configure all newly-constructed Sessions.
+    """
+    # HACK: We should consider modifying urllib3 to support this as an explicit module-level variable.
```
sigmavirus24: This comment doesn't make sense.

requests_toolbelt/_compat.py
```diff
@@ -18,11 +18,13 @@
     from requests.packages.urllib3 import fields
     from requests.packages.urllib3 import filepost
     from requests.packages.urllib3 import poolmanager
+    from requests.packages.urllib3.util import timeout
```
sigmavirus24: This fallback needs to be its own try/except block because it's causing the normal tests to fail.

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +53,75 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
         super(AppEngineAdapter, self).__init__(self, *args, **kwargs)
```
sigmavirus24: This needs a check that raises an exception for versions of requests < 2.10.0

requests_toolbelt/adapters/appengine.py
```diff
@@ -5,31 +5,40 @@
...
 import requests
 from requests import adapters
+from requests import sessions
 
 from .. import exceptions as exc
 from .._compat import gaecontrib
+from .._compat import timeout
 
+"""
+There are two ways to use this library.
```
sigmavirus24: This needs to document that it was added in v0.6.0 of the toolbelt and that this is a *submodule*.

```rst
.. versionadded:: 0.6.0
```

requests_toolbelt/adapters/appengine.py
```diff
@@ -44,6 +53,75 @@ def __init__(self, validate_certificate=True, *args, **kwargs):
...
+
+def monkeypatch():
+    """Sets up all Sessions to use AppEngineAdapter by default.
+
+    If you don't want to deal with constructing and configuring your own Sessions,
+    or if you use libraries that call the requests API directly (ie requests.post),
+    then you may prefer to monkeypatch and auto-configure all newly-constructed Sessions.
+    """
+    # HACK: We should consider modifying urllib3 to support this as an explicit module-level variable.
+    sessions.HTTPAdapter = AppEngineAdapter
```
sigmavirus24: This function also needs to check requests' version before doing anything.

requests_toolbelt/_compat.py
sigmavirus24: Also this fallback will need to be careful. If we do it unconditionally we can still break existing environments without a new enough requests+urllib3. We will need to do:

```py
try:
   from requests.packages.urllib3.util import timeout
except ImportError:
   try:
       from urllib3.util import timeout
   except ImportError:
       timeout = None
```

Alternatively, we can check if the build number is a certain value and then conditionally attempt to import this with a default value of `None`.

requests_toolbelt/_compat.py
```diff
@@ -18,11 +18,13 @@
...
 except ImportError:
     from urllib3 import connection
     from urllib3 import fields
     from urllib3 import filepost
     from urllib3 import poolmanager
+    from urllib3.util import timeout
 
 if requests.__build__ < 0x020800:
     gaecontrib = None
```
sigmavirus24: Ah we already have something like that here. We also need to update the version of urllib3 here because we're only going to properly support this for versions `>= 0x021000`

sigmavirus24:
Thank you @seawolf42 and @jonathan-s for testing this for us.

requests_toolbelt/_compat.py
jonathan-s: Out of curiousity, where do you find the build information. When I do `requests.__build__` on the latest version of requests (2.9.1) I get `133377`. Also curious what __build__ signifies.

requests_toolbelt/adapters/appengine.py
mikelambert: I'm curious, why does it need the check? Is this how we would check that https://github.com/shazow/urllib3/pull/763 had been merged properly into requests library?

requests_toolbelt/_compat.py
mikelambert: @jonathan-s '%0x' % 133377 == '20901'

mikelambert:
Okay, should have made all the changes you've requested. I believe all failing tests are now unrelated to my changes. And it will be harder to test and verify this code now (like was done earlier up-thread), because there is no 2.10.0 requests library yet. :)

requests_toolbelt/adapters/appengine.py
sigmavirus24: Yes, that is why we need that check.

sigmavirus24:
Could you rebase this please?

mikelambert:
Rebased.

sigmavirus24:
Thanks @mikelambert

mikelambert:
Thanks! Curious, what's the next step here? Push requests to release 2.10.0 with the latest urllib3, and then after you can integrate feature/gae into master?

sigmavirus24:
Please don't push requests to do anything. Speaking as a maintainer there, we're already aware this is a desirable feature.

Because of the version checking that takes place in the code now, we can actually release the toolbelt when it's ready. I just need to merge (at least) 2 PRs before performing a release here.

PR #120, opened 2015-12-17
Ensure proxies use correct TLS version.

Lukasa:
Credit for this work goes to [Quovo](https://www.quovo.com).

In the process of investigating a bug, I discovered that proxies take a different code path through the adapter: specifically, they don't use the poolmanager initiated via `init_poolmanager`, but instead use the `proxy_manager_for` function to build themselves a new ProxyManager. This meant they didn't get the correct TLS version from the `SSLAdapter`, and could lead to mysterious discrepancies between the with-and-without proxy case.

This change rectifies that problem.

@sigmavirus24 I'm not really sure how to test this with Betamax. Thoughts?

### DIFF
--- a/requests_toolbelt/adapters/ssl.py
+++ b/requests_toolbelt/adapters/ssl.py
@@ -48,3 +48,7 @@ class SSLAdapter(HTTPAdapter):
             maxsize=maxsize,
             block=block,
             ssl_version=self.ssl_version)
+
+    def proxy_manager_for(self, *args, **kwargs):
+        kwargs['ssl_version'] = self.ssl_version
+        return super(SSLAdapter, self).proxy_manager_for(*args, **kwargs)
--- a/requests_toolbelt/adapters/ssl.py
+++ b/requests_toolbelt/adapters/ssl.py
@@ -9,6 +9,8 @@ in this blog post:
 https://lukasa.co.uk/2013/01/Choosing_SSL_Version_In_Requests/
 
 """
+import requests
+
 from requests.adapters import HTTPAdapter
 
 from .._compat import poolmanager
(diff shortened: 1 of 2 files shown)


sigmavirus24:
There really isn't a way to use betamax to test this. Unit level tests would be better.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Lukasa:
@sigmavirus24 Hmm. How would you like this tested? I assume we just need to verify that `super` is called with the correct arguments...should I just patch `super` out?

sigmavirus24:
Er, what about patching out the ProxyManger from urllib and ensuring we do the right things with it when we call that method?

Lukasa:
Yeah, that could work. I'll try to restrict the scope of the test to make sure it doesn't break too much in the face of requests going all crazy.

Lukasa:
Done. =)

Lukasa:
Ok, so this test fails with early versions of requests v2 because we hadn't factored out the ProxyManager initialisation logic. In this case, I'm inclined to skip the test primarily because there is actually nothing we can do for requests versions that old: the code just doesn't make it possible for us to correctly set the ProxyManager up.

Are you ok with me skipping the test for the older environments, and then adding a note in the docstring mentioning this limitation?

sigmavirus24:
Works for me.
-- 
Sent from my Android device with K-9 Mail. Please excuse my brevity.

Lukasa:
Alrighty, done. =)

requests_toolbelt/adapters/ssl.py
```diff
@@ -48,3 +54,7 @@ def init_poolmanager(self, connections, maxsize, block=False):
             maxsize=maxsize,
             block=block,
             ssl_version=self.ssl_version)
+
+    def proxy_manager_for(self, *args, **kwargs):
+        kwargs['ssl_version'] = self.ssl_version
```
sigmavirus24: So we presently have https://github.com/sigmavirus24/requests-toolbelt/blob/master/requests_toolbelt/adapters/socket_options.py#L49 in the socket options adapter.

This method *will* cause problems on anything <2.4.0, right? Since it is *possible* that people are already using  this with a version older than 2.4.0, we should endeavor not to break them. 

If I misunderstood your comment about skipping older versions of requests than this can be safely ignored probably.

Lukasa: This method will not cause problems on 2.4.0 because the method it's overriding doesn't exist. =)

Lukasa: Oh wait no that's not true. On some versions it does. Good idea, I'll make the definition of this method conditional.

Lukasa: Done.

sigmavirus24:
:+1: Just waiting for travis to confirm my suspicions. :)

PR #127, opened 2016-01-13
Ignore non-digest challenges in proxy digest auth

rdn32:
If HTTPProxyDigestAuth finds itself looking at a non-digest proxy
authentication challenge then it should ignore it rather than fall
over trying to make sense of the challenge's content.


Lukasa:
The functionality here seems reasonable to me, though I know that @sigmavirus24 is going to have some comments re testing.

tests/test_auth.py
```diff
@@ -75,3 +77,54 @@ def test_handle_407_header_basic(self, extract_cookies_to_jar, proxy_auth_call):
         guess_auth.handle_407(r, *kwargs)
 
         proxy_auth_call.assert_called_with(req)
+
+
+class ProxyAuthHeaderMatcher(betamax.BaseMatcher):
```
pcreech: Tests already exist for HTTPProxyDigestAuth in /tests/test_proxy_digest_auth.py.  It might be beneficial to move these to that file

pcreech:
Functionality looks good to me, not the best with testing logic yet.
perhaps @sigmavirus24 can give that a glance over?

The failed checks appear to be related to an update in pep8, unrelated to these changes

tests/test_auth.py
rdn32: Oops - I didn't notice that file. I'll hold off moving the tests in case more radical changes to them are about to be suggested.

sigmavirus24:
Yeah if you can update this to do the following:

- Remove the betamax integration tests. Betamax doesn't work at a low-enough level to be useful for this.
- Add unit tests to the file that @pcreech created.

Meanwhile I'll update the toolbelt to fix the pep8 issues. Thanks @rdn32

sigmavirus24:
Nevermind, @pcreech already updated the toolbelt for that pep8 issue.

rdn32:
I made the changes to the tests. I also renamed the existing test method in TestProxyDigestAuth to something more specific to what it is testing, and moved some set-up code that wasn't relevant to the new test methods into it from setUp(). I hope that's OK.

tests/test_proxy_digest_auth.py
```diff
@@ -15,26 +16,88 @@ def setUp(self):
...
 
+    def test_no_challenge(self):
+        """Tests that a response containing no authentication challenge is
+        left alone.
+        """
+        connection = MockConnection()
+        first_response = connection.make_response(self.prepared_request)
+        first_response.status_code = 404
+
+        assert not self.auth.last_nonce
```
sigmavirus24: `assert not` is a bit ... vague. Could this be a bit more specific, e.g.,

```py
assert self.auth.last_nonce is None
```

That's much more explicit.

tests/test_proxy_digest_auth.py
```diff
@@ -15,26 +16,88 @@ def setUp(self):
...
         """
+        self.auth.last_nonce = "bH3FVAAAAAAg74rL3X8AAI3CyBAAAAAA"
+        self.auth.chal = {
+            'nonce': self.auth.last_nonce,
+            'realm': 'testrealm@host.org',
+            'qop': 'auth'
+        }
+
         # prepared_request headers should be clear before calling auth
         assert not self.prepared_request.headers.get('Proxy-Authorization')
```
sigmavirus24: I realize I didn't catch that here. Would you mind updating this too?

tests/test_proxy_digest_auth.py
```diff
@@ -15,26 +16,88 @@ def setUp(self):
...
         self.auth(self.prepared_request)
         assert self.prepared_request.headers.get('Proxy-Authorization')
```
sigmavirus24: Likewise can this be

```py
assert self.prepared_request.headers.get('Proxy-Authorization') is not None
```

sigmavirus24:
I'm truly sorry for the long review cycles @rdn32. I've been very busy lately.

PR #132, opened 2016-01-26
Check that long_description is valid

Ivoz:
[Rendering is currently broken](https://pypi.python.org/pypi/requests-toolbelt)

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -31,9 +31,11 @@ commands = flake8 {posargs} requests_toolbelt
 [testenv:docs]
 deps =
     sphinx>=1.3.0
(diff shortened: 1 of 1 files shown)


sigmavirus24:
Correct. I'd rather this be a separate test env. If you'd also like to fix the setup.py to not read the `LICENSE` file into the `license` parameter but instead use the correct classifiers, that'd be great.

Ivoz:
IMHO they work well together, given they're both related, colloquially check for "did we write rst correctly", and I sorta think it's wasteful spinning up another build job just to run it.

PR #137, opened 2016-03-04
Don't actually allow pip to install Requests 3.0.0

untitaker:
I assume that this max-version was introduced because Requests 3 could
break requests-toolbelt.


Lukasa:
Yes, that does seem reasonable doesn't it! @sigmavirus24, I just want you to confirm that this isn't deliberate, though I'm almost certain it isn't, so I won't merge this myself.

sigmavirus24:
Thanks @untitaker!

PR #138, opened 2016-04-21
Improvements to stream_response_to_file()

rcoup:
Problem: API client library saving large responses via `requests_toolbelt.downloadutils.stream.stream_response_to_file()` wants to enable simple progress reporting via a callback.

Changes:

1. ~~add `callback` parameter to `stream_response_to_file()`, which is a function that takes two arguments `(bytes_read, total_bytes)`. The function will be called at least twice (at 0 bytes and at the end). `total_bytes` can be `None` if there's no `Content-Length` header on the response.~~
2. add optional `chunksize` parameter to `stream_response_to_file()`, ~~and increase the default chunk size from 512 bytes to 64KB. This matches the chunk size used in `dowloadutils.tee`. This should improve performance significantly for large responses.~~
3. extract the filename/header/file-object parsing into a separate function
4. add support for passing a directory to save into using the Content-Disposition filename (rather than always using the current working directory)
5. only use the last file-part of Content-Disposition headers: eg. `Content-Disposition: attachment; filename=/etc/passwd`
6. prevent overwriting existing files using arbitrary filenames from the server. This is obviously backwards incompatible for apps that expect to overwrite the same file every time.

* [x] Tests
* [x] Docs

### DIFF
--- a/requests_toolbelt/downloadutils/stream.py
+++ b/requests_toolbelt/downloadutils/stream.py
@@ -20,7 +20,8 @@ def _get_filename(content_disposition):
     for match in _OPTION_HEADER_PIECE_RE.finditer(content_disposition):
         k, v = match.groups()
         if k == 'filename':
-            return v
+            # ignore any directory paths in the filename
+            return os.path.split(v)[1]
     return None
 
 
@@ -51,9 +52,11 @@ def get_download_file_path(response, path):
         # fully qualified file path
         filepath = path
     else:
-        response_filename = _get_filename(response.headers.get('content-disposition', ''))
+        response_filename = _get_filename(
+            response.headers.get('content-disposition', '')
+        )
         if not response_filename:
-            raise exc.StreamingError('No filename given to stream response to.')
+            raise exc.StreamingError('No filename given to stream response to')
 
         if path_is_dir:
             # directory to download to
@@ -112,6 +115,9 @@ def stream_response_to_file(response, path=None, chunksize=_DEFAULT_CHUNKSIZE):
         r = requests.get(url, stream=True)
         filename = stream.stream_response_to_file(r, path='myfile')
 
+    If the calculated download file path already exists, this function will
+    raise a StreamingError.
+
     Instead, if you want to manage the file object yourself, you need to
     provide either a :class:`io.BytesIO` object or a file opened with the
     `'b'` flag. See the two examples below for more details.
@@ -143,7 +149,8 @@ def stream_response_to_file(response, path=None, chunksize=_DEFAULT_CHUNKSIZE):
     :param path: *(optional)*, Either a string with the path to the location
         to save the response content, or a file-like object expecting bytes.
     :type path: :class:`str`, or object with a :meth:`write`
-    :param int chunksize: (optional), Size of chunk to attempt to stream (default 512B).
+    :param int chunksize: (optional), Size of chunk to attempt to stream
+        (default 512B).
     :returns: The name of the file, if one can be determined, else None
     :rtype: str
     :raises: :class:`requests_toolbelt.exceptions.StreamingError`
@@ -157,6 +164,8 @@ def stream_response_to_file(response, path=None, chunksize=_DEFAULT_CHUNKSIZE):
         filename = getattr(fd, 'name', None)
     else:
         filename = get_download_file_path(response, path)
+        if os.path.exists(filename):
+            raise exc.StreamingError("File already exists: %s" % filename)
         fd = open(filename, 'wb')
 
     for chunk in response.iter_content(chunk_size=chunksize):
--- a/requests_toolbelt/downloadutils/stream.py
+++ b/requests_toolbelt/downloadutils/stream.py
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 """Utilities for dealing with streamed requests."""
 import collections
+import os.path
 import re
 
 from .. import exceptions as exc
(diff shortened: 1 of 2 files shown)


requests_toolbelt/downloadutils/stream.py
```diff
@@ -12,6 +12,7 @@
     r';\s*(%s|[^\s;=]+)\s*(?:=\s*(%s|[^;]+))?\s*' % (_QUOTED_STRING_RE,
                                                      _QUOTED_STRING_RE)
 )
+_DEFAULT_CHUNKSIZE = 65536
```
sigmavirus24: Why would you change the default chunksize from 512 to this?

requests_toolbelt/downloadutils/stream.py
```diff
@@ -122,8 +138,19 @@ def stream_response_to_file(response, path=None):
             )
         fd = open(filename, 'wb')
 
-    for chunk in response.iter_content(chunk_size=512):
+    try:
+        bytes_total = int(response.headers.get('content-length', None))
+    except TypeError:
+        bytes_total = None  # not specified or invalid
```
sigmavirus24: If it's not specified than it was probably sent using `Transfer-Encoding: chunked` which is not invalid.

This should be done differently, *if* at all.

sigmavirus24:
`stream_response_to_file` does not need to know about a callback. Here's why:

```py
class MonitoringFile(object):
    def __init__(self, filename, mode, total_bytes):
        self.filename = filename
        self.file = open(filename, mode)
        self.total_bytes = total_bytes
        self.bytes_written = 0

    def write(self, data):
        if not data:
            return
        self.bytes_written += len(data)
        self.file.write(data)

response = requests.get(url, stream=True)
total_bytes = get_download_length(response)
fd = MonitoringFile('my-streamed-file.txt', 'wb', total_bytes)
stream_response_to_file(response, fd)
```

sigmavirus24:
That said, I'm fine with allowing the chunksize to be configurable, but not to changing it without good justification.

requests_toolbelt/downloadutils/stream.py
rcoup: logic: 512 bytes is very low for most download where you care about streaming to a file, and `tee` already uses 64KB.

requests_toolbelt/downloadutils/stream.py
rcoup: There's plenty of responses that (quite rightly) don't include a Content-Length. But yeah, can easily rework into an `if` statement and a separate `int()` cast .

rcoup:
> stream_response_to_file does not need to know about a callback. Here's why:

Sure, it doesn't _need_ to, just trying to provide a helper. The wider use case for me is wrapping `stream_response_to_file()` as a method in another library, but keeping the useful functionality around being able to pass in:
* a file-like object
* a file path
* or have the filename parsed from `Content-Disposition`
And provide a simple progress callback option for a user of my library.

If we did the `MonitoringFile` approach then we'd need to reimplement the `path` handling at the outer level and after that there's not a lot left of it :)

requests_toolbelt/downloadutils/stream.py
sigmavirus24: Except that tee has a much different use case. In this case, we're dealing with a streamed response from requests and there have been issues in the past where we found 512 to be a good chunksize. Making this configurable will allow you to default it to whatever you'd like in your wrapper.

sigmavirus24:
> If we did the MonitoringFile approach then we'd need to reimplement the path handling at the outer level and after that there's not a lot left of it :)

Which suggests that we need to refactor that into a separate helper function in this module. That makes it more testable and also allows others to use it when implementing extra features.

I'm fairly certain this function doesn't need the ability to handle callbacks. @Lukasa, thoughts?

rcoup:
Update - I have reworked into several changes as discussed above. New strategy is better?

Each is a separate commit, so hopefully straightforward to review that way.

requests_toolbelt/downloadutils/stream.py
```diff
@@ -12,17 +14,72 @@
...
+    """
+    pre_opened = False
+    fd = None
+    filename = None
+    if path:
+        if isinstance(getattr(path, 'write', None), collections.Callable):
+            pre_opened = True
+            fd = path
+            filename = getattr(fd, 'name', None)
+        elif os.path.isdir(path):
```
sigmavirus24: Why not more simply:

```py
elif not os.path.isdir(path):
    # If the user specified a directory, let the 
    filename = path
    if os.path.exists(filename):
        raise
    if os.path.exists(filename):
        raise exc.StreamingError(...)
    fd = open(filename, 'wb')
```

That said, why are we now going to start allowing people to specify the path to a directory now?

sigmavirus24:
I'm sorry. I wasn't very clear. I want a new *public* function that will simply return the filename to be used from a Response. This will allow you to call that first and then call `os.path.join(directory, filename)` and pass that to `stream_response_to_file`.

rcoup:
> I want a...

No problem, will take another stab at it :)

> That said, why are we now going to start allowing people to specify the path to a directory now?

Downloading to arbitrary filenames in `cwd` seems like a bad idea to encourage. I mean, we can clearly document it but it's still not great. 

Downloading to a specified directory seems tidier, and doesn't require `os.chdir()` (which changes it for all threads in a MT app, and still isn't available as a context manager in stdlib).

Personally I'd suggest eventually deprecating `path=None` to avoid unexpected behaviour, and suggest users use `path=os.curdir` if that's what they _actually_ want.

rcoup:
Revisiting an earlier point:

> stream_response_to_file does not need to know about a callback. Here's why:

Would you prefer including that example in the docs, or bundling a basic implementation ala `MultipartEncoderMonitor`?

rcoup:
ok, implemented as `get_download_file_path()` (improved naming welcome), and moved more of the logic back into `stream_response_to_file()`. 

Re-included the dangerous pathname and existing file stuff as a separate commit.

sigmavirus24:
Triggering a new merged build

rcoup:
Fixed the pyflakes errors.

PR #139, opened 2016-04-21
Include RTD theme for Sphinx 1.4.0+

rcoup:
With Sphinx 1.4, the RTD theme became optional, and needs to be included as a dependency in `tox.ini`

Otherwise when you run `tox -e docs` you get:
```
...
docs runtests: commands[0] | sphinx-build -E -c docs -b html docs/ docs/_build/html
Running Sphinx v1.4.1

Theme error:
sphinx_rtd_theme is no longer a hard dependency since version 1.4.0. Please install it manually.
ERROR: InvocationError: '/Users/rcoup/code/requests-toolbelt/.tox/docs/bin/sphinx-build -E -c docs -b html docs/ docs/_build/html'
______________________________________________________________________________________ summary ______________________________________________________________________________________
ERROR:   docs: commands failed
```

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -99,9 +99,15 @@ pygments_style = 'sphinx'
 
 # -- Options for HTML output ----------------------------------------------
 
+on_rtd = os.environ.get('READTHEDOCS', None) == 'True'
+if not on_rtd:  # only import and set the theme if we're building docs locally
+    import sphinx_rtd_theme
+    html_theme = 'sphinx_rtd_theme'
+    html_theme_path = [sphinx_rtd_theme.get_html_theme_path()]
+
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
-html_theme = 'sphinx_rtd_theme'
+#html_theme = 'alabaster'
 
 # Theme options are theme-specific and customize the look and feel of a theme
 # further.  For a list of options available for each theme, see the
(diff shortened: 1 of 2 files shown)


sigmavirus24:
This is not complete: https://github.com/snide/sphinx_rtd_theme#via-package

sigmavirus24:
Also https://github.com/snide/sphinx_rtd_theme#using-this-theme-locally-then-building-on-read-the-docs

rcoup:
Dunno, I'm a sphinx/RTD novice :) This allowed the docs to build as per the instructions in CONTRIBUTING and I could see them.

sigmavirus24:
> Dunno, I'm a sphinx/RTD novice :) This allowed the docs to build as per the instructions in CONTRIBUTING and I could see them.

Were they themed properly? Either way, I trust that they documented this appropriately so I'd rather we follow that documentation.

rcoup:
The actual change for 1.4 just [pulled it out of Sphinx' setup.py](https://github.com/sphinx-doc/sphinx/pull/2087) -- so the [theme notes](https://github.com/snide/sphinx_rtd_theme#using-this-theme-locally-then-building-on-read-the-docs) suggests it doesn't work *already* on RTD?

I've added their doc'd approach, and it builds and tests ok in via `tox`

sigmavirus24:
Thanks! :tada:

PR #140, opened 2016-04-28
Add pip install note

austinhyde:
Because otherwise the only place this is stated is in the `setup.py`.


sigmavirus24:
Thank you @austinhyde

PR #148, opened 2016-05-10
Add support for BasedSession

jaraco:
Following on the [suggestion in requests](https://github.com/kennethreitz/requests/issues/2554#issuecomment-109341010), I've implemented the suggestion. Please let me know if there's anything more you'd like to see to get this accepted. I use this technique all the time and find the code copy/pasted in several places and I'd like to consolidate it somewhere.

Thanks.

### DIFF
--- a/docs/index.rst
+++ b/docs/index.rst
@@ -36,6 +36,7 @@ Full Documentation
     dumputils
     formdata
     exceptions
+    sessions
     threading
     uploading-data
     user-agent
--- a/docs/sessions.rst
+++ b/docs/sessions.rst
@@ -1,0 +1,9 @@
+.. _sessions:
+
+Specialized Sessions
+====================
+
+The toolbelt provides specialized session classes.
+
+.. automodule:: requests_toolbelt.sessions
+    :members:
--- a/requests_toolbelt/_compat.py
+++ b/requests_toolbelt/_compat.py
@@ -44,10 +44,11 @@ PY3 = sys.version_info > (3, 0)
 
 if PY3:
     import queue
-    from urllib.parse import urlencode
+    from urllib.parse import urlencode, urljoin
 else:
     import Queue as queue
     from urllib import urlencode
+    from urlparse import urljoin
 
 try:
     basestring = basestring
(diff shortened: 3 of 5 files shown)


jaraco:
Damn. Failures in Python 2.7 suggest `requests.Session` is an old-style class.

Lukasa:
`requests.Session` is not an old-style class.

jaraco:
> requests.Session is not an old-style class.

Oh! Of course! It's the mix-in that's old style. My bad.

sigmavirus24:
@jaraco could you explain the benefit of having the `Based` mix-in and the `BasedSession` class separate like you do? I'm not sure I see the benefit at the moment.

jaraco:
@sigmavirus24 My intention there was to make the underlying functionality available to other requests.Session subclasses that other users may have in their libs/apps. I'm happy to fold that together or make it private at your preference.

sigmavirus24:
I'm just having trouble thinking of a case where they wouldn't just subclass `BasedSession` if they're already sub-classing a session. I think I must be missing something.

docs/index.rst
```diff
@@ -36,6 +36,7 @@ Full Documentation
     dumputils
     formdata
     exceptions
+    sessions
```
sigmavirus24: I think you forgot to add this file too.

jaraco:
I've updated the code to simplify the implementation to just one subclass, and added the missing docs. Anything else to consider before acceptance?

sigmavirus24:
@jaraco just time. I'll probably get around to this tomorrow and add some more docs to it when I merge it.

Thanks! 🎉

jaraco:
I just went to use this today and found to my chagrin it still isn't available. Anything I can do to speed the adoption?

sigmavirus24:
Nope. This reminder is good enough. I'm going to put some time into this project tonight and cut a release. I'm so sorry I lost track of this.

jaraco:
Thank you.

sigmavirus24:
0.7.0 is up on PyPI with this in it.

foxx:
Thanks @jaraco and @sigmavirus24 for getting this done, I completely forgot to send a PR!

PR #153, opened 2016-07-07
Proxy support for SourceAddressAdapter

dys:
Using proxy_manager_for to pass source_address to the ProxyManager.

### DIFF
--- a/requests_toolbelt/adapters/source.py
+++ b/requests_toolbelt/adapters/source.py
@@ -63,4 +63,5 @@ class SourceAddressAdapter(HTTPAdapter):
 
     def proxy_manager_for(self, *args, **kwargs):
         kwargs['source_address'] = self.source_address
-        return super(SourceAddressAdapter, self).proxy_manager_for(*args, **kwargs)
+        return super(SourceAddressAdapter, self).proxy_manager_for(
+            *args, **kwargs)
--- a/requests_toolbelt/adapters/source.py
+++ b/requests_toolbelt/adapters/source.py
@@ -60,3 +60,7 @@ class SourceAddressAdapter(HTTPAdapter):
             maxsize=maxsize,
             block=block,
             source_address=self.source_address)
+
+    def proxy_manager_for(self, *args, **kwargs):
+        kwargs['source_address'] = self.source_address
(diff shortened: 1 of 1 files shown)


requests_toolbelt/adapters/source.py
```diff
@@ -60,3 +60,7 @@ def init_poolmanager(self, connections, maxsize, block=False):
             maxsize=maxsize,
             block=block,
             source_address=self.source_address)
+
+    def proxy_manager_for(self, *args, **kwargs):
+        kwargs['source_address'] = self.source_address
```
sigmavirus24: What if instead we do

```py
kwargs.setdefault('source_address', self.source_address)
```

I suppose it's possible that's already being sent and we wouldn't want to forcibly override it? I'm not sure.

sigmavirus24:
Also, the tests are failing on style compliance: https://travis-ci.org/sigmavirus24/requests-toolbelt/jobs/143071491#L165

dys:
Ha! 3 characters over. Doh. The class name is longer than SSLAdapter. I'll see if I can find a way to make it a bit shorter.

RE: setdefault, SSLAdapter is doing something similar with ssl_version. Would it be good to change there too for consistency?

requests_toolbelt/adapters/source.py
```diff
@@ -60,3 +60,7 @@ def init_poolmanager(self, connections, maxsize, block=False):
...
+        return super(SourceAddressAdapter, self).proxy_manager_for(*args, **kwargs)
```
sigmavirus24: Try:

```py
return super(SourceAddressAdapter, self).proxy_manager_for(
    *args, **kwargs
)
```

sigmavirus24:
> RE: setdefault, SSLAdapter is doing something similar with ssl_version. Would it be good to change there too for consistency?

So that should be a different pull request if we decide to do it at all. I'd like @Lukasa to weigh in before deciding to do this or not.

sigmavirus24:
Fair enough. @dys if you just fix up the style issue, I'm happy to merge this. :)

sigmavirus24:
Thanks @dys!

PR #157, opened 2016-07-15
Allow users to access an IP address via HTTPS

huoneefi:
Requesting a IP address directly like this fails with an Exception:

    requests.get(url="https://93.184.216.34", headers={"Host": "example.org"})

SSL certificate verification doesn't respect the explicitly set host header
and fails with "hostname doesn't match".

This commit adds a SNIAdapter which sets the host name used for verification
based on the HTTP host header.

It can be used like so:

    from requests_toolbelt import SNIAdapter
    s = requests.Session()
    s.mount('https://', SNIAdapter())
    s.get("https://93.184.216.34", headers={"Host": "example.org"})

see: https://github.com/kennethreitz/requests/issues/3403

### DIFF
--- a/requests_toolbelt/__init__.py
+++ b/requests_toolbelt/__init__.py
@@ -9,7 +9,7 @@ See http://toolbelt.rtfd.org/ for documentation
 :license: Apache v2.0, see LICENSE for more details
 """
 
-from .adapters import SSLAdapter, SourceAddressAdapter
+from .adapters import SSLAdapter, SourceAddressAdapter, SNIAdapter
 from .auth.guess import GuessAuth
 from .multipart import (
     MultipartEncoder, MultipartEncoderMonitor, MultipartDecoder,
@@ -27,7 +27,7 @@ __version_info__ = tuple(int(i) for i in __version__.split('.'))
 
 __all__ = [
     'GuessAuth', 'MultipartEncoder', 'MultipartEncoderMonitor',
-    'MultipartDecoder', 'SSLAdapter', 'SourceAddressAdapter',
+    'MultipartDecoder', 'SSLAdapter', 'SourceAddressAdapter', 'SNIAdapter',
     'StreamingIterator', 'user_agent', 'ImproperBodyPartContentException',
     'NonMultipartContentTypeException', '__title__', '__authors__',
     '__license__', '__copyright__', '__version__', '__version_info__',
--- a/requests_toolbelt/adapters/sni.py
+++ b/requests_toolbelt/adapters/sni.py
@@ -1,0 +1,45 @@
+# -*- coding: utf-8 -*-
+"""
+requests_toolbelt.sni_adapter
+=============================
+
+This file contains an implementation of the SNIAdapter
+"""
+
+from requests.adapters import HTTPAdapter
+
+
+class SNIAdapter(HTTPAdapter):
+    """
+    A HTTPS Adapter for Python Requests that sets the hostname for certificate
+    verification based on the host header.
+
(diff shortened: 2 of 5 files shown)


davidfischer:
The host header is set automatically based on the requested hostname. Since you didn't request a hostname (you requested an IP) it didn't get set automatically. For example, if you try to access github.com by IP address (https://192.30.253.112/) you will get a certificate mismatch. In your example, instead of `s.get("https://93.184.216.34", headers={"Host": "example.org"})`, you could just do `s.get("https://example.org")` assuming that example.org actually resolved to 93.184.216.34.

huoneefi:
Yes, you're right. But if you can't resolve DNS and must request the IP address directly, you get a certificate mismatch despite explicitly setting the HTTP host header. This Adapter solve this problem by setting the expected host used for certificate validation based on the HTTP host header.

See kennethreitz/requests#3403 and the documentation and tests for an example.

requests_toolbelt/__init__.py
```diff
@@ -27,7 +27,7 @@
 
 __all__ = [
     'GuessAuth', 'MultipartEncoder', 'MultipartEncoderMonitor',
-    'MultipartDecoder', 'SSLAdapter', 'SourceAddressAdapter',
+    'MultipartDecoder', 'SSLAdapter', 'SourceAddressAdapter', 'SNIAdapter',
```
sigmavirus24: We've stopped adding things to the top level module. We plan to remove all of these in 1.0.0. I'll remove this when I merge it.

requests_toolbelt/adapters/sni.py
```diff
@@ -0,0 +1,45 @@
...
+        for header in request.headers:
+            if header.lower() == "host":
+                host = request.headers[header]
+                break
+
+        if host:
+            self.poolmanager.connection_pool_kw["assert_hostname"] = host
+        elif "assert_hostname" in self.poolmanager.connection_pool_kw:
+            # an assert_hostname from a previous request may have been left
+            del self.poolmanager.connection_pool_kw["assert_hostname"]
```
sigmavirus24: If you mount this and never actually send a Host header, you'll never be able to send a request:

```py
~/s/requests-toolbelt git:master ❯❯❯ python
Python 2.7.11 (default, Jun 15 2016, 20:12:08)
[GCC 4.2.1 Compatible Apple LLVM 7.3.0 (clang-703.0.31)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> a = {}
>>> del a['assert_hostname']
  File "<stdin>", line 1, in <module>
KeyError: 'assert_hostname'
>>>
```

tests/test_sniadapter.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def test_case_insensitive_header(self):
+        r = self.session.get("https://93.184.216.34",
+                             headers={"hOSt": "example.org"})
+        assert r.status_code == 200
+
+    def test_plain_requests(self):
+        # test whether the reason for this adapter remains
+        # (may be implemented into requests in the future)
+        with self.assertRaises(requests.exceptions.SSLError):
```
sigmavirus24: We use py.test, this should use `pytest.raises`.

sigmavirus24:
Superseded by #158.

requests_toolbelt/adapters/sni.py
Lukasa: This note is extremely important because, in regular use, the Host header is added by `httplib`: that is, at a *lower* level than this one.

PR #158, opened 2016-07-22
SSL Adapter that peforms verification based on the Host Header

sigmavirus24:
Hey @Lukasa 

Can you give this a look over? :eyes:

This was originally #157 but I've pulled it and made a bunch of changes. Just let me know what you think.


Lukasa:
LGTM. =)

PR #165, opened 2016-10-13
Also monkeypatch adapters.HTTPAdapter

JelteF:
Should fix https://github.com/GetStream/stream-python/issues/43


sigmavirus24:
Thanks @JelteF and @Lukasa!

PR #169, opened 2016-12-09
Disable cert validation when monkeypatching AppEngine

bendemaree:
We have a use case for this in which, due to an issue with AppEngine's `dev_appserver.py`, certain certificates will fail validation; more on that issue [here](https://code.google.com/p/googleappengine/issues/detail?id=13154). The only real workaround when using the emulator is to disable certificate validation.

Because this is a large code base, in which requests is used both explicitly and within third-party libraries, we are using the monkeypatch; however, to my eye there's no way to set the `validate_certificate` kwarg in the constructor for `AppEngineAdapter` to `True` when monkeypatching; if we could, it would serve our use case by allowing global disabling of certificate validation when a local environment is detected.

This creates a new `AppEngineInsecureAdapter` in which cert validation is disabled by default and adds a kwarg to `monkeypatch()` to trigger its use. Initially I tried to simply make the monkeypatch set `partial(AppEngineAdapter, validate_certificate=False)` but that was a bit naive as the adapter is not just instantiated.

### DIFF
--- a/docs/adapters.rst
+++ b/docs/adapters.rst
@@ -53,6 +53,25 @@ ways to take advantage of this support at the moment:
 
        appengine.monkeypatch()
 
+If you should need to disable certification validation when monkeypatching (to
+force third-party libraries that use requests to not validate certificates, if
+they do not provide API surface to do so, for example), you can disable it:
+
+   .. code-block:: python
+
+       from requests_toolbelt.adapters import appengine
+       appengine.monkeypatch(validate_certificate=False)
+
+   .. warning::
+
+       If ``validate_certificate`` is ``False``, the monkeypatched adapter
+       will *not* validate certificates. This effectively sets the
+       ``validate_certificate`` argument to urlfetch.Fetch() to ``False``. You
+       should avoid using this wherever possible. Details can be found in the
+       `documentation for urlfetch.Fetch()`__.
+
+       __ https://cloud.google.com/appengine/docs/python/refdocs/google.appengine.api.urlfetch
+
 .. autoclass:: requests_toolbelt.adapters.appengine.AppEngineAdapter
 
 FingerprintAdapter
--- a/docs/adapters.rst
+++ b/docs/adapters.rst
@@ -53,6 +53,8 @@ ways to take advantage of this support at the moment:
 
        appengine.monkeypatch()
 
+.. _insecure_appengine:
+
 If you should need to disable certificate validation when monkeypatching (to
 force third-party libraries that use Requests to not validate certificates, if
 they do not provide API surface to do so, for example), you can disable it:
@@ -68,9 +70,9 @@ they do not provide API surface to do so, for example), you can disable it:
        will *not* validate certificates. This effectively sets the
        ``validate_certificate`` argument to urlfetch.Fetch() to ``False``. You
        should avoid using this wherever possible. Details can be found in the
-       `documentation for urlfetch.Fetch()`__.
+       `documentation for urlfetch.Fetch()`_.
 
-       __ https://cloud.google.com/appengine/docs/python/refdocs/google.appengine.api.urlfetch
+       .. _documentation for urlfetch.Fetch(): https://cloud.google.com/appengine/docs/python/refdocs/google.appengine.api.urlfetch
 
 .. autoclass:: requests_toolbelt.adapters.appengine.AppEngineAdapter
 
@@ -150,7 +152,7 @@ SourceAddressAdapter
 
 .. versionadded:: 0.3.0
 
-The :class:`~requests_toolbelt.adapters.source.SourceAddressAdapter` allows a 
+The :class:`~requests_toolbelt.adapters.source.SourceAddressAdapter` allows a
 user to specify a source address for their connnection.
 
 .. autoclass:: requests_toolbelt.adapters.source.SourceAddressAdapter
--- a/docs/adapters.rst
+++ b/docs/adapters.rst
@@ -53,8 +53,8 @@ ways to take advantage of this support at the moment:
 
        appengine.monkeypatch()
 
-If you should need to disable certification validation when monkeypatching (to
-force third-party libraries that use requests to not validate certificates, if
+If you should need to disable certificate validation when monkeypatching (to
+force third-party libraries that use Requests to not validate certificates, if
 they do not provide API surface to do so, for example), you can disable it:
 
    .. code-block:: python
--- a/requests_toolbelt/adapters/appengine.py
+++ b/requests_toolbelt/adapters/appengine.py
@@ -30,24 +30,8 @@ There are two ways to use this library:
 
 which will ensure all requests.Session objects use AppEngineAdapter properly.
 
-If you should need to disable certificate validation when monkeypatching (to
-force third-party libraries that use Requests to not validate certificates, if
-they do not provide API surface to do so, for example), you can disable it:
-
-   .. code-block:: python
-
-       >>> from requests_toolbelt.adapters import appengine
-       >>> appengine.monkeypatch(validate_certificate=False)
-
-   .. warning::
-
-       If ``validate_certificate`` is ``False``, the monkeypatched adapter
-       will *not* validate certificates. This effectively sets the
-       ``validate_certificate`` argument to urlfetch.Fetch() to ``False``. You
-       should avoid using this wherever possible. Details can be found in the
-       `documentation for urlfetch.Fetch()`__.
-
-       __ https://cloud.google.com/appengine/docs/python/refdocs/google.appengine.api.urlfetch  # noqa
+You are also able to :ref:`disable certificate validation <insecure_appengine>`
+when monkey-patching.
 """
 import requests
 import warnings
@@ -80,8 +64,16 @@ class AppEngineAdapter(adapters.HTTPAdapter):
 
 
 class InsecureAppEngineAdapter(AppEngineAdapter):
-    """A variant of the the transport adapter for Requests to use urllib3's GAE
-    support that does not validate certificates by default. Use with caution!
+    """An always-insecure GAE adapter for Requests.
+
+    This is a variant of the the transport adapter for Requests to use
+    urllib3's GAE support that does not validate certificates. Use with
+    caution!
+
+    .. note::
+        The ``validate_certificate`` keyword argument will not be honored here
+        and is not part of the signature because we always force it to
+        ``False``.
 
     See :class:`AppEngineAdapter` for further details.
     """
@@ -91,7 +83,7 @@ class InsecureAppEngineAdapter(AppEngineAdapter):
             warnings.warn("Certificate validation cannot be specified on the "
                           "InsecureAppEngineAdapter, but was present. This "
                           "will be ignored and certificate validation will "
-                          "remain off.", exc.IgnoringCertificateValidation)
+                          "remain off.", exc.IgnoringGAECertificateValidation)
 
         super(InsecureAppEngineAdapter, self).__init__(
             validate_certificate=False, *args, **kwargs)
@@ -177,9 +169,8 @@ def monkeypatch(validate_certificate=True):
     # HACK: We should consider modifying urllib3 to support this cleanly,
     # so that we can set a module-level variable in the sessions module,
     # instead of overriding an imported HTTPAdapter as is done here.
-    if validate_certificate:
-        adapter = AppEngineAdapter
-    else:
+    adapter = AppEngineAdapter
+    if not validate_certificate:
         adapter = InsecureAppEngineAdapter
 
     sessions.HTTPAdapter = adapter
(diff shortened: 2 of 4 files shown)


requests_toolbelt/adapters/appengine.py
```diff
@@ -59,6 +78,22 @@ def init_poolmanager(self, connections, maxsize, block=False):
         self.poolmanager = _AppEnginePoolManager(self._validate_certificate)
 
 
+class AppEngineInsecureAdapter(adapters.HTTPAdapter):
+    """A variant of the the transport adapter for Requests to use urllib3's GAE
+    support that does not validate certificates by default. Use with caution!
+
+    See `.AppEngineAdapter` for further details.
```
sigmavirus24:
```
`.AppEngineAdapter`
```

Isn't valid ReStructuredText it needs to be

```rst
:class:`AppEngineAdapter`
```

requests_toolbelt/adapters/appengine.py
```diff
@@ -59,6 +78,22 @@ def init_poolmanager(self, connections, maxsize, block=False):
         self.poolmanager = _AppEnginePoolManager(self._validate_certificate)
 
 
+class AppEngineInsecureAdapter(adapters.HTTPAdapter):
```
sigmavirus24: Better question, why did you copy an entire class definition instead of doing

```py
class InsecureAppEngineAdapter(AppEngineAdapter):

    def __init__(self, validate_certificate=False, *args, **kwargs):
        super(InsecureAppEngineAdapter, self).__init__(validate_certificate, *args, **kwargs)
```

The AppEngineAdapter already creates the `_AppEnginePoolManager`.

sigmavirus24:
> Initially I tried to simply make the monkeypatch set partial(AppEngineAdapter, validate_certificate=False) but that was a bit naive as the adapter is not just instantiated.

Can you explain this a bit further? What problems did you find with this approach?

requests_toolbelt/adapters/appengine.py
sigmavirus24: No worries.

bendemaree:
I ran into [this check](https://github.com/kennethreitz/requests/blob/master/requests/adapters.py#L117) which resulted in

```
  File "requests_toolbelt/adapters/appengine.py", line 76, in __init__
    super(AppEngineAdapter, self).__init__(*args, **kwargs)
  File "equests/adapters.py", line 117, in __init__
    super(HTTPAdapter, self).__init__()
TypeError: super() argument 1 must be type, not functools.partial
```

sigmavirus24:
I suspect the monkey-patch is patching the wrong thing then (or we're mis-using imports in the toolbelt).

sigmavirus24:
What happens if you delete [this line](https://github.com/sigmavirus24/requests-toolbelt/blob/7679220dc272e0c824eed8fa12d6adc702e21bcb/requests_toolbelt/adapters/appengine.py#L138)?

bendemaree:
I'll give it a go in a few hours here. The possibility of the patching being off might explain some confusing errors I got in my initial attempt to write some test around this as well. Thanks for the pointers!

sigmavirus24:
@bendemaree excellent investigation!

So looking into #165 again, I have mixed feelings about it, but it seems to be completely valid. So I guess we're back to the method that you're currently employing in this pull request (having two different adapter).

I think this is best done like so:

```py
class InsecureAppEngineAdapter(AppEngineAdapter):

    def __init__(self, *args, **kwargs):  # Note: there's no validate_certificate parameter here
        # Maybe ensure "validate_certificate" is not in **kwargs
        super(InsecureAppEngineAdapter, self).__init__(validate_certificate=False, *args, **kwargs)
```

And then we should key the monkey-patch as you're currently doing. Thanks for sticking with this @bendemaree

bendemaree:
@sigmavirus24 My pleasure, thanks for all you do!

I added some tests around this, and the monkeypatch itself, now that everything is making sense. I also added a warning if the user instantiates `InsecureAppEngineAdapter` manually with `validate_certificate` specified (which I'm hoping will never happen!). I went on the precedent of `warnings` already being in use; however, I copy-pasted from `adapters/socket_options.py` and I think [the usage there](https://github.com/sigmavirus24/requests-toolbelt/blob/master/requests_toolbelt/adapters/socket_options.py#L46) might be incorrect (message then class unless first arg is an instance).

Lastly--I took a hint from your notes and renamed it to `InsecureAppEngineAdapter`...I had a suspicion that was the more intuitive name myself.

docs/adapters.rst
```diff
@@ -53,6 +53,25 @@ ways to take advantage of this support at the moment:
...
+       from requests_toolbelt.adapters import appengine
+       appengine.monkeypatch(validate_certificate=False)
+
+   .. warning::
+
+       If ``validate_certificate`` is ``False``, the monkeypatched adapter
+       will *not* validate certificates. This effectively sets the
+       ``validate_certificate`` argument to urlfetch.Fetch() to ``False``. You
+       should avoid using this wherever possible. Details can be found in the
+       `documentation for urlfetch.Fetch()`__.
```
sigmavirus24: This should end in a single underscore

docs/adapters.rst
```diff
@@ -53,6 +53,25 @@ ways to take advantage of this support at the moment:
...
+
+       __ https://cloud.google.com/appengine/docs/python/refdocs/google.appengine.api.urlfetch
```
sigmavirus24: This should be

```rst
.. _documentation for urlfetch.Fetch(): <url>
```

docs/adapters.rst
```diff
@@ -53,6 +53,25 @@ ways to take advantage of this support at the moment:
 
        appengine.monkeypatch()
 
+If you should need to disable certificate validation when monkeypatching (to
```
sigmavirus24: Perhaps we should add a link ref here, e.g.,

```
.. _insecure_appengine:

If you should need to disable ...
```

requests_toolbelt/adapters/appengine.py
```diff
@@ -29,8 +29,28 @@
...
+
+   .. warning::
+
+       If ``validate_certificate`` is ``False``, the monkeypatched adapter
+       will *not* validate certificates. This effectively sets the
+       ``validate_certificate`` argument to urlfetch.Fetch() to ``False``. You
+       should avoid using this wherever possible. Details can be found in the
+       `documentation for urlfetch.Fetch()`__.
+
+       __ https://cloud.google.com/appengine/docs/python/refdocs/google.appengine.api.urlfetch  # noqa
```
sigmavirus24: Same thing as above, although this looks duplicated from there. Perhaps remove the duplication and add a reference to it, e.g.,

```rst
You are also able to :ref:`disable certificate validation <insecure_appengine>` when monkey-patching.
```

requests_toolbelt/adapters/appengine.py
```diff
@@ -59,6 +79,24 @@ def init_poolmanager(self, connections, maxsize, block=False):
         self.poolmanager = _AppEnginePoolManager(self._validate_certificate)
 
 
+class InsecureAppEngineAdapter(AppEngineAdapter):
+    """A variant of the the transport adapter for Requests to use urllib3's GAE
+    support that does not validate certificates by default. Use with caution!
+
+    See :class:`AppEngineAdapter` for further details.
```
sigmavirus24: We should probably add to the docstring:

```rst

.. note::

    The ``validate_certificate`` keyword argument will not be honored here and is not part of the signature because we always force it to ``False``.
```

requests_toolbelt/adapters/appengine.py
```diff
@@ -59,6 +79,24 @@ def init_poolmanager(self, connections, maxsize, block=False):
         self.poolmanager = _AppEnginePoolManager(self._validate_certificate)
 
 
+class InsecureAppEngineAdapter(AppEngineAdapter):
+    """A variant of the the transport adapter for Requests to use urllib3's GAE
```
sigmavirus24: I know we don't make pydocstyle a required check in Travis, but please try to follow PEP-0257. For example,

```
An always-insecure GAE adapter for Requests.

This is a variant of the AppEngineAdapter provided by this module that will **never** validate certificates by default. **Use with caution!**
```

requests_toolbelt/adapters/appengine.py
```diff
@@ -123,19 +161,29 @@ def urlopen(self, method, url, body=None, headers=None, retries=None,
...
+        If ``validate_certificate`` is ``False``, certification validation will
+        effectively be disabled for all requests.
     """
     _check_version()
     # HACK: We should consider modifying urllib3 to support this cleanly,
     # so that we can set a module-level variable in the sessions module,
     # instead of overriding an imported HTTPAdapter as is done here.
-    sessions.HTTPAdapter = AppEngineAdapter
-    adapters.HTTPAdapter = AppEngineAdapter
+    if validate_certificate:
```
sigmavirus24: Since there's other review feedback here, what if we do:

```py
adapter = AppEngineAdapter
if not validate_certificate:
    adapter = InsecureAppEngineAdapter
```

requests_toolbelt/exceptions.py
```diff
@@ -23,3 +23,12 @@ class RequestsVersionTooOld(Warning):
     this warning to the user.
     """
     pass
+
+
+class IgnoringCertificateValidation(Warning):
```
sigmavirus24: Should probably make this mildly more specific, e.g.,

```
class IgnoringGAECertificateValidation(Warning):
```

requests_toolbelt/adapters/appengine.py
bendemaree: Whoops, sloppy; thanks for the catch. Fixed in the warning docstring as well.

docs/adapters.rst
bendemaree: It's almost like I'm just winging the RST... 😬

requests_toolbelt/exceptions.py
```diff
@@ -17,17 +17,20 @@ class VersionMismatchError(Exception):
 
 
 class RequestsVersionTooOld(Warning):
-    """Used to indiciate that the Requests version is too old.
+    """Used to indicate that the Requests version is too old.
```
sigmavirus24: Nice catch!

sigmavirus24:
This looks excellent! Thank you!

bendemaree:
Rock on, thanks for dealing with my poor doc skills. And, stay warm down there! ❄️

sigmavirus24:
> And, stay warm down there

I'll do my best!

> thanks for dealing with my poor doc skills

Your first attempts are 100x better than what I used to write. Documentation writing is something I struggle with constantly. Most of my edits were syntax related (which is the easiest part of ReStructuredText for me).

bendemaree:
@sigmavirus24 Thanks for the kind words!

Looks like there's a little more work to do here...I went to plugin this in to the project and got more infinite recursion errors. I then installed it into the toy app above; same thing. I went back again to #165 and re-thought things...can you give your reasoning why the change was valid? I don't see any way that `AppEngineAdapter` can inherit from `HTTPAdapater`...and then _replace_ it as `requests.adapters.HTTPAdapter`. As soon as `AppEngineAdapter` is instantiated, it calls super, which, due to the monkeypatch, is itself.

IMO the original request to patch might be invalid, because in that case a specific adapter is being requested. The solution originally posed in the PR seems more valid (importing the adapter assigned to `requests.session`).

sigmavirus24:
So the only reason I think the original report *might* be valid is that Requests doesn't really document *one* way to import an use an HTTPAdapter yourself. So if you look at the code-base that was running into problems with that, they were importing HTTPAdapter from `requests.adapters` which is just as valid as importing it from `requests.sessions`.

Unfortunately @Lukasa is on holiday, so we can't really gather his feedback right now. Let me see if I can find someone else who has opinions about this kind of stuff. Maybe @jonparrott

jonparrott:
I'm on my way out as well, I'll see if I can take a closer look later today
before I disappear.

sigmavirus24:
Thanks @jonparrott

jonparrott:
> So the only reason I think the original report might be valid is that Requests doesn't really document one way to import an use an HTTPAdapter yourself. So if you look at the code-base that was running into problems with that, they were importing HTTPAdapter from requests.adapters which is just as valid as importing it from requests.sessions.

Yeah this import aliasing is going to cause all kinds of fun stuff when you try to replace something with its own subclass.

You could do some jiggerypockery where you create a reference to the original `HTTPAdapter` class before patching and ensure that `AppEngineAdapter` inherits from that.

All of that seems pretty gross, though.

sigmavirus24:
> All of that seems pretty gross, though.

I agree 110%. :)

I didn't like the monkey-patching idea at all, but it seemed necessary. Now it seems to be a nightmare. :)

jonparrott:
Maybe we should just ask requests to make it easier to do this without
monkey patching.

sigmavirus24:
@jonparrott the problem is that requests-toolbelt aims to support as many versions of requests as possible, so that would only work for people who keep up-to-date with the latest and greatest.

bendemaree:
@sigmavirus24 @jonparrott have opened a PR with a potential fix; PTAL! 😄

PR #170, opened 2016-12-20
AppEngine monkeypatching MRO fix

bendemaree:
This works around some of the issues found in #169.

With my esteemed colleague @Bogdanp, we determined the issue that arises when monkeypatching is not actually the inheritance relationship set up between these classes, or the import aliasing; it's when the MRO is re-calculated for the monkeypatched class. This behavior can be reproduced with this trivial example:

```python
class Foo(object):
    def __init__(self):
        super(Foo, self).__init__()


class Bar(Foo):
    def __init__(self):
        super(Bar, self).__init__()


Foo = Bar
Bar()
```

The way @Bogdanp devised to work around this is to add an intermediate class that makes the MRO be calculated properly; there's some more detail in the docstring and commit around this.

### DIFF
--- a/requests_toolbelt/adapters/appengine.py
+++ b/requests_toolbelt/adapters/appengine.py
@@ -43,7 +43,24 @@ from .._compat import gaecontrib
 from .._compat import timeout
 
 
-class AppEngineAdapter(adapters.HTTPAdapter):
+class AppEngineMROHack(adapters.HTTPAdapter):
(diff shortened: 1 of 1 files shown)


Bogdanp:
An in-depth explanation of the hack is available here: https://www.youtube.com/watch?v=eYlJH81dSiw

sigmavirus24:
Thanks @bendemaree and @Bogdanp

PR #175, opened 2017-02-10
handle last boundary without CRLF

mtrbean:
The spec doesn't require CRLP after the close-delimiter https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html

### DIFF
--- a/requests_toolbelt/multipart/decoder.py
+++ b/requests_toolbelt/multipart/decoder.py
@@ -143,7 +143,10 @@ class MultipartDecoder(object):
             return BodyPart(fixed, self.encoding)
 
         def test_part(part):
-            return part != b'' and part != b'\r\n' and part[:4] != b'--\r\n'
+            return (part != b'' and
+                    part != b'\r\n' and
+                    part[:4] != b'--\r\n' and
+                    part != b'--')
 
         parts = self.content.split(b''.join((b'\r\n', boundary)))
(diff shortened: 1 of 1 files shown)


requests_toolbelt/multipart/decoder.py
```diff
@@ -143,7 +143,7 @@ def body_part(part):
             return BodyPart(fixed, self.encoding)
 
         def test_part(part):
-            return part != b'' and part != b'\r\n' and part[:4] != b'--\r\n'
+            return part != b'' and part != b'\r\n' and part[:4] != b'--\r\n' and part != '--'
```
sigmavirus24: Let's rewrite this to be

```py
return (part != b'' and
        part != b'\r\n' and
        part[:4] != b'--\r\n' and
        part != b'--')
```

The last `b'--'` is actually import given we're a single-source Python 2 and 3 compatible library.

mtrbean:
Thanks! I've made the changes that you request. The server that I'm working with doesn't send the last CRLF.

Are you able release 0.7.1 with this change?

sigmavirus24:
I'm not sure it would be a `0.7.1` and not `0.8.0`. I need to check the log. Also, I only release things on Mondays or Tuesdays given the predominant corporate usage of my libraries.

sigmavirus24:
Looking over the log, it would be 0.7.1

PR #176, opened 2017-02-13
Fixed 'get_encodings_from_content' incompatible with Python3.

### LINKED ISSUE #172: utils.get_encodings_from_content doesn't work with Python 3
Since the parameter `content` is byte string in Python 3, the re variable `find_charset`/`find_pragma`/`find_xml` won't match against `content` and complains:
`cannot use a string pattern on a bytes-like object`
It will be better to use explicit byte- prefix for regular expressions.

<bountysource-plugin>
---
Want to back this issue? **[Post a bounty on it!](https://www.bountysource.com/issues/40614752-utils-get_encodings_from_content-doesn-t-work-with-python-3?utm_campaign=plugin&utm_content=tracker%2F418367&utm_medium=issues&utm_source=github)** We accept bounties via [Bountysource](https://www.bountysource.com/?utm_campaign=plugin&utm_content=tracker%2F418367&utm_medium=issues&utm_source=github).
</bountysource-plugin>

adoal:
This patch fixes #172  :)

### DIFF
--- a/requests_toolbelt/utils/deprecated.py
+++ b/requests_toolbelt/utils/deprecated.py
@@ -4,6 +4,16 @@ import re
 
 from requests import utils
 
+from .._compat import PY3 as _PY3
+
+_re_charset = r'<meta.*?charset=["\']*(.+?)["\'>]'
+_re_pragma = r'<meta.*?content=["\']*;?charset=(.+?)["\'>]'
+_re_xml = r'^<\?xml.*?encoding=["\']*(.+?)["\'>]'
+
+if _PY3:
+    _re_charset = _re_charset.encode('ascii')
+    _re_pragma = _re_pragma.encode('ascii')
+    _re_xml = _re_xml.encode('ascii')
 
 def get_encodings_from_content(content):
     """Return encodings from given content string.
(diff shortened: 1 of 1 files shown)


requests_toolbelt/utils/deprecated.py
```diff
@@ -4,6 +4,17 @@
 
 from requests import utils
 
+from .._compat import PY3 as _PY3
+
+_re_charset = r'<meta.*?charset=["\']*(.+?)["\'>]'
```
sigmavirus24: You can actually write each of these as

```
br'...'
```

We're always expecting bytes and `r'...' == br'...'` on Python 2 and `r'...'.encode('ascii') == br'...'` on Python 3. =)

requests_toolbelt/utils/deprecated.py
```diff
@@ -20,15 +31,15 @@ def get_encodings_from_content(content):
     :type content: bytes
     """
     find_charset = re.compile(
```
sigmavirus24: Can we move the compile calls up to the module level while we're here? Thanks!

requests_toolbelt/utils/deprecated.py
```diff
@@ -6,15 +6,17 @@
 
 from .._compat import PY3 as _PY3
```
sigmavirus24: Can you remove this too? We don't need it anymore.

adoal:
It's fixed. Thank you for your patience with a newbie :)

requests_toolbelt/utils/deprecated.py
adoal: It's fixed. Thank you for your patience with a newbie :)

sigmavirus24:
Thanks for taking the time to fix this @adoal!

PR #178, opened 2017-02-21
Update deprecated.py : get_encodings_from_content

luoxzhg:
correct the return type of get_encoding_from_content in Python3, from a list of bytes to a list of str.

### DIFF
--- a/requests_toolbelt/utils/deprecated.py
+++ b/requests_toolbelt/utils/deprecated.py
@@ -1,7 +1,7 @@
 # -*- coding: utf-8 -*-
 """A collection of functions deprecated in requests.utils."""
 import re
-
+import sys
 from requests import utils
 
 find_charset = re.compile(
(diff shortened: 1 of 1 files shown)


requests_toolbelt/utils/deprecated.py
```diff
@@ -32,7 +32,8 @@ def get_encodings_from_content(content):
     :type content: bytes
```
sigmavirus24: We should instead just document here that we're returning a list of bytes objects. For example:

```rst
:returns: Detected encodings in the content as byte-strings.
:rtype: list(bytes)
```

On Python 2 ```str is bytes``` so it's accurate, while on Python 3, ```bytes is bytes``` so the documentation will be more accurate.

I realize that we accidentally changed the return type on Python 3 for this function, but if we accept this change, we'll do the opposite on Python 2 by returning unicode there.

luoxzhg: I have fixed this problem.
If we do not change it, get_unicode_from_response will be break in Python3. Because, bytes.decode(encoding) method or str(bytes, encoding) can not accept a bytes encoding string, it must be a str.

sigmavirus24:
Thanks @luoxzhg!

PR #188, opened 2017-05-29
Added fix for OSX TCPKeepAliveAdapter.

irothschild:


requests_toolbelt/adapters/socket_options.py
```diff
@@ -103,13 +104,23 @@ def __init__(self, **kwargs):
...
-        # set it conditionally.
+        # NOTE(Ian): OSX does not have these constants defined, so we
+        # set them conditionally.
+        if getattr(socket, 'TCP_KEEPINTVL', None) is not None:
+            socket_options += [(socket.IPPROTO_TCP, socket.TCP_KEEPINTVL,
+                                interval)]
+        elif sys.platform == 'darwin':
+            # On OSX, TCP_KEEPALIVE from netinet/tcp.h is not exported
+            # by python's socket module
+            # TCP_KEEPALIVE = 0x10
```
sigmavirus24: Is it never exported or is it not exported on some versions of Python? If the latter, I'd prefer:

```
TCP_KEEPALIVE = getattr(socket, 'TCP_KEEPALIVE', 0x10)
```

rather than hard-coding this.

irothschild: As far as I know it's not exported on any Python version but your change is good.

sigmavirus24:
✨ 🍰 ✨

Issue #194, opened 2017-10-02
Threading documentation should say: "num_processes" not "num_threads"


afhaque:
Auto-Generated Documentation for Threading appears to have an incorrect key.

![image](https://user-images.githubusercontent.com/5855843/31063488-497e835a-a701-11e7-8c72-7bcd4934c96c.png)

The argument correctly displays: "num_processes" as the key in the class, but lists the same argument incorrectly as "num_threads" in the description. 

http://toolbelt.readthedocs.io/en/latest/threading.html

PR #195, opened 2017-10-02
Corrected Pool documentation

afhaque:
See: https://github.com/requests/toolbelt/issues/194


sigmavirus24:
✨ You're awesome! Thanks for catching and correcting this! ✨

PR #197, opened 2017-10-30
Coerce proxy URL to bytes before dumping request

johnhawes:
The proxy URL is returned as a string from _build_request_path().
This throws a TypeError when it is added to the prefix and method
in _dump_request_data(). A fix is to coerce the proxy URL to bytes
before returning it.


sigmavirus24:
Well caught! Thanks!

PR #200, opened 2017-11-21
Fix bug/199, enabling support for contenttype with capital letters.

### LINKED ISSUE #199: Multipart decoder is casesensitive in evaluating string "multipart"
Encountered a instance at a customers where the server used towards the multipart decoder returned the string "Multipart/Related", this failing the multipart decoder; since the at present only accepts lowercase "multipart".

Ref. /requests_toolbelt/multipart/decoder.py

```
def _find_boundary(self):
...
        if mimetype.split('/')[0] != 'multipart':
            raise NonMultipartContentTypeException(
                "Unexpected mimetype in content-type: '{0}'".format(mimetype)
            )
...
```

That should have been
```
def _find_boundary(self):
...
        if mimetype.split('/')[0].lower() != 'multipart':
            raise NonMultipartContentTypeException(
                "Unexpected mimetype in content-type: '{0}'".format(mimetype)
            )
...
```
Seems like adding '.lower()' fixes the problem, without adding further complications.

Kefevs:
Some servers return Content-type with cap ex 'Multipart/Related'.
This breaks the present implementation.

Closes #199

### DIFF
--- a/requests_toolbelt/multipart/decoder.py
+++ b/requests_toolbelt/multipart/decoder.py
@@ -115,7 +115,7 @@ class MultipartDecoder(object):
     def _find_boundary(self):
         ct_info = tuple(x.strip() for x in self.content_type.split(';'))
         mimetype = ct_info[0]
-        if mimetype.split('/')[0] != 'multipart':
+        if mimetype.split('/')[0].lower() != 'multipart':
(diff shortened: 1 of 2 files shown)


sigmavirus24:
Thanks for fixing this!

PR #201, opened 2017-12-29
Prevent `to_string` from modifying `MultipartEncoder`'s buffer

adamtheturtle:
Previously, `MultipartEncoder`'s behaviour was changed after using the
`to_string` method. This was confusing behaviour. Now, the buffer is
copied internally, so the `MultipartEncoder` remains the same after
using `to_string`.

### DIFF
--- a/requests_toolbelt/multipart/encoder.py
+++ b/requests_toolbelt/multipart/encoder.py
@@ -8,6 +8,7 @@ This holds all of the implementation details of the MultipartEncoder
 
 """
 import contextlib
+import copy
 import io
 import os
 from uuid import uuid4
@@ -279,16 +280,16 @@ class MultipartEncoder(object):
             streaming or reading data from the encoder, this method will only
             return whatever data is left in the encoder.
 
-        .. note::
-
-            This method affects the internal state of the encoder. Calling
-            this method will exhaust the encoder.
-
         :returns: the multipart message
         :rtype: bytes
         """
+        if self.finished:
+            buffer_copy = copy.copy(self._buffer)
+            return buffer_copy.read()
 
-        return self.read()
+        self._load(-1)
+        buffer_copy = copy.copy(self._buffer)
+        return buffer_copy.read()
 
     def read(self, size=-1):
         """Read data from the streaming encoder.
--- a/AUTHORS.rst
+++ b/AUTHORS.rst
@@ -41,3 +41,5 @@ Patches and Suggestions
 - Mike Lambert (@mikelambert)
 
 - Ryan Barrett (https://snarfed.org/)
+
+- Adam Dangoor (@adamdangoor)
--- a/tests/test_multipart_encoder.py
+++ b/tests/test_multipart_encoder.py
@@ -98,6 +98,9 @@ class TestMultipartEncoder(unittest.TestCase):
             '--this-is-a-boundary--\r\n'
         ).encode()
 
+    def test_to_string_twice(self):
+        assert self.instance.to_string() == self.instance.to_string()
+
     def test_content_type(self):
         expected = 'multipart/form-data; boundary=this-is-a-boundary'
         assert self.instance.content_type == expected


requests_toolbelt/multipart/encoder.py
```diff
@@ -279,16 +280,16 @@ def to_string(self):
             streaming or reading data from the encoder, this method will only
```
adamtheturtle: I built the documentation but cannot find where this is rendered.

sigmavirus24:
Hi there,

Thanks for sending this pull request. I'm wondering if you could share a bit more about your use case that is motivating this change. I'd like to better understand the reasoning behind this change before discussing it further.

Cheers,
Ian

adamtheturtle:
Hi Ian,

My code makes an "Image Recognition Query" to Vuforia's API, documented at https://library.vuforia.com/articles/Solution/How-To-Perform-an-Image-Recognition-Query.

I need to have the file data twice, once to construct the `Authorization` header and once to make the request.

My code currently looks like this:

```python
image_content = high_quality_image.read()
date = rfc_1123_date()
request_path = '/v1/query'
url = urljoin('https://cloudreco.vuforia.com', request_path)
files = {'image': ('image.jpeg', image_content, 'image/jpeg')}

multipart_encoded_data = MultipartEncoder(fields=files)

# We copy the encoded data because `to_string` mutates the object.
content = copy.deepcopy(multipart_encoded_data).to_string()

authorization_string = authorization_header(
    access_key='client_access_key',
    secret_key='client_secret_key',
    method=POST,
    content=content,
    # Note that this is not the actual Content-Type header value sent.
    content_type='multipart/form-data',
    date=date,
    request_path=request_path,
)

headers = {
    'Authorization': authorization_string,
    'Date': date,
    'Content-Type': multipart_encoded_data.content_type,
}

response = requests.request(
    method=POST,
    url=url,
    headers=headers,
    data=multipart_encoded_data,
)

assert response.status_code == codes.OK
```

With this proposed change, I would not need to use `copy.deepcopy`.

Without the proposed change, if I did not use `copy.deepcopy`, the Vuforia API would give me a confusing `504`. This took me some time to debug.

sigmavirus24:
Okay that's actually super helpful to understand. Thanks.

At the moment, I feel hesitant about accepting this change for a few reasons.

First and foremost, the `MultipartEncoder` is meant to be a tool that people use when they have large files whose contents they're trying to upload without exhausting their RAM or using too much memory. That purpose and constraint led to some implementation decisions that effect whether or not this can *actually* be idempotent. On this line: https://github.com/requests/toolbelt/blob/16f294f927266ef74bc7e39e0793e59f13a023bc/requests_toolbelt/multipart/encoder.py#L183 we *always* truncate our buffer. We do this to eliminate content from memory so we're not going to use a terrible amount of memory. In fact, we will likely trim ~8 KB of data from that buffer each time.

Based on the code you provided, it looks as if you're reading the entire file contents in no matter what (`image_content = high_quality_image.read()`) and then constructing the fields like so `files = {'image': ('image.jpeg', image_content, 'image/jpeg')}`. That along with the fact that you explicitly do `content = copy.deepcopy(multipart_encoded_data).to_string()` to pass the content into your `Authorization` function makes me wonder why you don't just also pass `content` to `requests`. I *suspect* that I'm missing something. Have I gotten anything wrong?

adamtheturtle:
Thank you so much for your consideration and detailed response.

The issue was simply a misunderstanding on my part. Passing `content` to requests works and stops me having to use `copy`.

I will say that this was a matter of confusion, and that I would not have been confused with this change - I didn't expect `to_string` to change `multipart_encoded_data`. However, I'm happy to accept that this is worth the trade-off for memory.

adamtheturtle:
This means that I could also use:

```python
encoded_data, encoded_data_content_type = encode_multipart_formdata(
    fields=files,
)
...
            'Content-Type': encoded_data_content_type,
...
```

sigmavirus24:
> I will say that this was a matter of confusion, and that I would not have been confused with this change - I didn't expect to_string to change multipart_encoded_data.

Yeah, that much is clear. I would love to hear how I could make this less confusing for others.

It seems like part of the problem is that we use `autoclass` here https://github.com/requests/toolbelt/blob/master/docs/uploading-data.rst#streaming-multipart-data-encoder and that doesn't display the documentation for `to_string` [here](http://toolbelt.readthedocs.io/en/latest/uploading-data.html#requests_toolbelt.multipart.encoder.MultipartEncoder)

We could update https://github.com/requests/toolbelt/blob/16f294f927266ef74bc7e39e0793e59f13a023bc/requests_toolbelt/multipart/encoder.py#L74 to include `.. automethod:: to_string` and `.. autoattribute:: content_type` so those show up in the documentation too. Does that sound good?

And yes, you can also just use `encode_multipart_formdata` from urllib3 to do this for yourself more directly. In fact, you could probably take advantage of building an AuthHandler for requests that would receive the Request object and iterate over that information for yourself.

adamtheturtle:
> ...to include .. automethod:: to_string and .. autoattribute:: content_type so those show up in the documentation too. Does that sound good?

Yes it does, but it would not have prevented me personally from being confused.
Admittedly, Vuforia's 504 response is not helpful, but I did not know that the issue was in my use of the encoder at first.

The only documentation that I had looked at was the README:

> Or, you can just create the string and examine the data:
> ```
> # Assuming `m` is one of the above
> m.to_string()  # Always returns unicode
> ```

Perhaps, a comment describing the side-effects here would be useful.

However, as a user, I don't think that documentation alone is enough to entirely eliminate confusion.

PR #208, opened 2018-02-01
Make sure SourceAddressAdapter can be pickled

benruijl:
This commit overrides `__getstate__` and `__setstate__` to store the source  address as part of the state. As a result, the `SourceAddressAdapter`  can be pickled correctly.

For more information, see: https://stackoverflow.com/questions/27711907/python-unpickle-a-object-with-a-class-instance-inside

### DIFF
--- a/requests_toolbelt/adapters/source.py
+++ b/requests_toolbelt/adapters/source.py
@@ -54,6 +54,15 @@ class SourceAddressAdapter(HTTPAdapter):
 
         super(SourceAddressAdapter, self).__init__(**kwargs)
 
+    def __getstate__(self):
+        state = super(SourceAddressAdapter, self).__getstate__()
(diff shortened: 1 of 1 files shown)


requests_toolbelt/adapters/source.py
```diff
@@ -54,6 +54,15 @@ def __init__(self, source_address, **kwargs):
 
         super(SourceAddressAdapter, self).__init__(**kwargs)
 
+    def __getstate__(self):
```
sigmavirus24: Requests defines [`__attrs__`](https://github.com/requests/requests/blob/8da118dfeb15989c47241d6ad6b7c429a34a51f3/requests/adapters.py#L108) which we could extend here, e.g.,

```py
__attrs__ = HTTPAdapter.__attrs__ + ['source_address']
```

PR #219, opened 2018-06-07
 Fix bug when MultipartEncoder is asked to encode zero parts

thauk-copperleaf:
Fixes error `AttributeError: 'NoneType' object has no attribute 'bytes_left_to_write'` at https://github.com/requests/toolbelt/blob/6d74296f0f267d9e06ba7828585dac61276d4e2b/requests_toolbelt/multipart/encoder.py#L193


tests/test_multipart_encoder.py
```diff
@@ -311,6 +311,12 @@ def test_accepts_custom_headers(self):
         output = m.read().decode('utf-8')
         assert output.index('X-My-Header: my-value\r\n') > 0
 
+    def test_no_parts(self):
+        fields = []
+        boundary = '--90967316f8404798963cce746a4f4ef9'
+        m = MultipartEncoder(fields=fields, boundary=boundary)
+        output = m.read().decode('utf-8')
+        assert output == '----90967316f8404798963cce746a4f4ef9--\r\n'
```
sigmavirus24: I'm not certain that writing anything here is correct behaviour, honestly. I wonder if we should set `self.finished = not self.parts` where `not []` would be `True`.

sigmavirus24: That's good enough for me.

PR #222, opened 2018-06-16
Added support for streaming multipart decoding

tzickel:
While this is feature complete, tests and documentation need to be added. This is mainly here to get feedback regarding the API and the possibility of putting this inside requests_toolbelt.

```python
r = requests.get('some multipart url', stream=True)
for part in MultipartStreamDecoder.from_response(r):
  print(part.headers)
  for stream in part:
    print(stream)
```


tzickel:
https://gist.github.com/tzickel/4a81503acdb843dab4f03cfe950e84f3

This is a benchmark for this code that shows a potential use case for it. You can mess with the data size and chunk size in the end and see the time differences between the versions (peak memory measurement is much more tricky and depends on your OS).

PR #224, opened 2018-06-29
Fix callback signature in docs to match the code. Closes #223

### LINKED ISSUE #223: MultipartEncoderMonitor's callback documentation is wrong
The callback docs say 
http://toolbelt.readthedocs.io/en/latest/uploading-data.html#requests_toolbelt.multipart.encoder.MultipartEncoderMonitor

```python
def callback(encoder, bytes_read):
```

same with the comments in the code:

https://github.com/requests/toolbelt/blob/master/requests_toolbelt/multipart/encoder.py#L342

but the actual callback only sends an instance of `self`
https://github.com/requests/toolbelt/blob/master/requests_toolbelt/multipart/encoder.py#L403

DavidVentura:


requests_toolbelt/multipart/encoder.py
```diff
@@ -339,7 +339,7 @@ class MultipartEncoderMonitor(object):
                                        MultipartEncoderMonitor)
         import requests
 
-        def callback(encoder, bytes_read):
+        def callback(encoder):
```
sigmavirus24: Should we rename this to `monitor` instead of `encoder` to make it yet clearer what instance you're actually getting?

requests_toolbelt/multipart/encoder.py
```diff
@@ -357,7 +357,7 @@ def callback(encoder, bytes_read):
         from requests_toolbelt import MultipartEncoderMonitor
         import requests
 
-        def callback(encoder, bytes_read):
+        def callback(encoder):
```
sigmavirus24: Same concern here

sigmavirus24:
Perfect! Thank you!

PR #227, opened 2018-10-05
Allow users to manipulate data before dumping

### LINKED ISSUE #226: Redacting sensitive headers when dumping request/response
In our project we have implemented dumping of request and response when debug-mode is enabled. Since this might end up in logfiles or other places where sensitive data is unwanted, we implemented a blacklist for sensitive headers.

Could something similar be implemented in the `dump_response` and `dump_all` methods here? Then we could get rid of our own implementation and just use toolbelt. Maybe make it optional?

Feel free to look at/borrow from our implementation here: https://github.com/fiaas/k8s/blob/master/k8s/client.py

mortenlj:
Dumping the request/response is useful for debugging, but that means
the data usually will end up in logfiles or other places. Since many
requests or responses will contain sensitive data (authentication
tokens, username, cookies etc), it is sometimes necessary to redact some
of that information before dumping it. This implementation provides a
way to supply your own sanitizer, or use the provided sanitizer which
will redact information in a number of known sensitive headers.

Closes #226

### DIFF
--- a/requests_toolbelt/utils/dump.py
+++ b/requests_toolbelt/utils/dump.py
@@ -3,7 +3,7 @@ import collections
 
 from requests import compat
 
-__all__ = ('dump_response', 'dump_all', 'Sanitizer', 'HeaderSanitizer')
+__all__ = ('dump_response', 'dump_all', 'NoopSanitizer', 'HeaderSanitizer')
 
 HTTP_VERSIONS = {
     9: b'0.9',
@@ -45,19 +45,26 @@ _PrefixSettings = collections.namedtuple('PrefixSettings',
 
 
 class Sanitizer(object):
-    """Performs no sanitation"""
+    REDACTED_VALUE = "#REDACTED#"
 
-    def request_header(self, name, value):
-        """Sanitize a request header
+    def _sanitize_headers(self, headers):
+        sanitized_headers = headers.copy()
+        for name in headers:
+            if self.should_sanitize_header(name):
+                sanitized_headers[name] = self.REDACTED_VALUE
+            if self.should_strip_header(name):
+                del sanitized_headers[name]
+        return sanitized_headers
 
-        :param name: The header name
-        :type name: :class:`requests.compat.basestring`
-        :param value: The header value
-        :type value: :class:`requests.compat.basestring`
-        :return: The value to dump for this header
-        :rtype: :class:`requests.compat.basestring`
+    def request_headers(self, headers):
+        """Sanitize the request headers
+
+        :param headers: The request headers
+        :type headers: :class:`requests.structures.CaseInsensitiveDict`
+        :return: A new headers object
+        :rtype: :class:`requests.structures.CaseInsensitiveDict`
         """
-        return value
+        return self._sanitize_headers(headers)
 
     def request_body(self, body):
         """Sanitize a request body
@@ -67,19 +74,17 @@ class Sanitizer(object):
         :return: The value to dump for the body
         :rtype: `bytes`
         """
-        return body
+        raise NotImplementedError()
 
-    def response_header(self, name, value):
-        """Sanitize a response header
+    def response_headers(self, headers):
+        """Sanitize the response headers
 
-        :param name: The header name
-        :type name: :class:`requests.compat.basestring`
-        :param value: The header value
-        :type value: :class:`requests.compat.basestring`
-        :return: The value to dump for this header
-        :rtype: :class:`requests.compat.basestring`
+        Modify the headers in place, removing or redacting values
+
+        :param headers: The response headers
+        :type headers: :class:`requests.structures.CaseInsensitiveDict`
         """
-        return value
+        return self._sanitize_headers(headers)
 
     def response_body(self, body):
         """Sanitize a request body
@@ -89,24 +94,40 @@ class Sanitizer(object):
         :return: The value to dump for the body
         :rtype: `bytes`
         """
+        raise NotImplementedError()
+
+    def should_sanitize_header(self, name):
+        raise NotImplementedError()
+
+    def should_strip_header(self, name):
+        raise NotImplementedError()
+
+
+class NoopSanitizer(Sanitizer):
+    """Performs no sanitation"""
+
+    def should_sanitize_header(self, name):
+        return False
+
+    def should_strip_header(self, name):
+        return False
+
+    def request_body(self, body):
+        return body
+
+    def response_body(self, body):
         return body
 
 
-class HeaderSanitizer(Sanitizer):
+class HeaderSanitizer(NoopSanitizer):
     """Redact the values of headers considered sensitive
 
     This will check all headers in both request and response against a list
     of sensitive headers (see :const:`SENSITIVE_HEADERS`), and redact the
     values to protect sensitive data.
     """
-    def _sanitize_headers(self, name, value):
-        """Redact the values of headers considered sensitive"""
-        if name.lower() in SENSITIVE_HEADERS:
-            return "#REDACTED#"
-        return value
-
-    request_header = _sanitize_headers
-    response_header = _sanitize_headers
+    def should_sanitize_header(self, name):
+        return name.lower() in SENSITIVE_HEADERS
 
 
 class PrefixSettings(_PrefixSettings):
@@ -152,7 +173,7 @@ def _dump_request_data(request, prefixes, bytearr, proxy_info=None,
     if proxy_info is None:
         proxy_info = {}
     if sanitizer is None:
-        sanitizer = Sanitizer()
+        sanitizer = NoopSanitizer()
 
     prefix = prefixes.request
     method = _coerce_to_bytes(proxy_info.pop('method', request.method))
@@ -166,8 +187,8 @@ def _dump_request_data(request, prefixes, bytearr, proxy_info=None,
     host_header = _coerce_to_bytes(headers.pop('Host', uri.netloc))
     bytearr.extend(prefix + b'Host: ' + host_header + b'\r\n')
 
-    for name, value in headers.items():
-        value = sanitizer.request_header(name, value)
+    sanitized_headers = sanitizer.request_headers(headers)
+    for name, value in sanitized_headers.items():
         bytearr.extend(prefix + _format_header(name, value))
 
     bytearr.extend(prefix + b'\r\n')
@@ -185,7 +206,7 @@ def _dump_request_data(request, prefixes, bytearr, proxy_info=None,
 
 def _dump_response_data(response, prefixes, bytearr, sanitizer=None):
     if sanitizer is None:
-        sanitizer = Sanitizer()
+        sanitizer = NoopSanitizer()
 
     prefix = prefixes.response
     # Let's interact almost entirely with urllib3's response
@@ -199,10 +220,9 @@ def _dump_response_data(response, prefixes, bytearr, sanitizer=None):
                    str(raw.status).encode('ascii') + b' ' +
                    _coerce_to_bytes(response.reason) + b'\r\n')
 
-    headers = raw.headers
-    for name in headers.keys():
-        for value in headers.getlist(name):
-            value = sanitizer.response_header(name, value)
+    sanitized_headers = sanitizer.response_headers(raw.headers)
+    for name in sanitized_headers.keys():
+        for value in sanitized_headers.getlist(name):
             bytearr.extend(prefix + _format_header(name, value))
 
     bytearr.extend(prefix + b'\r\n')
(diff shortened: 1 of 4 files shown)


requests_toolbelt/utils/dump.py
```diff
@@ -3,19 +3,116 @@
...
 
 HTTP_VERSIONS = {
     9: b'0.9',
     10: b'1.0',
     11: b'1.1',
 }
 
+#: List of sensitive headers copied from
+#: https://github.com/google/har-sanitizer
+SENSITIVE_HEADERS = {
```
sigmavirus24: I do not remember agreeing to wholesale import a list of headers to automagically sanitize. I thought we had agreed to merely allow users to provide a sanitizer, not give them a built-in one.

requests_toolbelt/utils/dump.py
```diff
@@ -115,12 +219,17 @@ def _coerce_to_bytes(data):
...
 
 def dump_response(response, request_prefix=b'< ', response_prefix=b'> ',
-                  data_array=None):
+                  data_array=None, sanitizer=NOOP_SANITIZER):
     """Dump a single request-response cycle's information.
 
     This will take a response object and dump only the data that requests can
     see for that single request-response cycle.
 
+    If the optional `sanitize` parameter is used, it should be an object that
```
sigmavirus24: This should be

```
If the optional ``sanitize`` parameter is used, it should be an object that
```

Note the double backticks. They're important because we're not using markdown.

requests_toolbelt/utils/dump.py
```diff
@@ -3,19 +3,116 @@
...
+}
+
 _PrefixSettings = collections.namedtuple('PrefixSettings',
                                          ['request', 'response'])
 
 
+class Sanitizer(object):
+    """Performs no sanitation"""
+
+    def request_header(self, name, value):
```
sigmavirus24: I believe that we should sanitize `headers` not individual pairs. How would someone, for example, strip a header name-value pair if they wanted to?

requests_toolbelt/utils/dump.py
```diff
@@ -3,19 +3,116 @@
...
+    "x-client-data",
+    "samlrequest",
+    "samlresponse"
+}
+
 _PrefixSettings = collections.namedtuple('PrefixSettings',
                                          ['request', 'response'])
 
 
+class Sanitizer(object):
```
sigmavirus24: There should be a sanitizer interface of sorts that people subclass and override methods without the methods defaulting to unsafe behaviours. Let's have this be a "base class" that raises `NotImplementedError()` exceptions and then an `NoopSanitizer` class that implements this behaviour.

requests_toolbelt/utils/dump.py
```diff
@@ -71,20 +169,24 @@ def _dump_request_data(request, prefixes, bytearr, proxy_info=None):
     bytearr.extend(prefix + b'Host: ' + host_header + b'\r\n')
 
     for name, value in headers.items():
+        value = sanitizer.request_header(name, value)
```
sigmavirus24: Since our sanitizer will be sanitizing the entire headers dictionary, let's move this outside the loop

requests_toolbelt/utils/dump.py
```diff
@@ -115,12 +219,17 @@ def _coerce_to_bytes(data):
 
 
 def dump_response(response, request_prefix=b'< ', response_prefix=b'> ',
-                  data_array=None):
+                  data_array=None, sanitizer=NOOP_SANITIZER):
```
sigmavirus24: Let's leave this as having a default of `None` and then instantiate a new `NoopSanitizer` if it is None. (Same for above.)

requests_toolbelt/utils/dump.py
```diff
@@ -153,18 +265,24 @@ def dump_response(response, request_prefix=b'< ', response_prefix=b'> ',
...
-                       proxy_info=proxy_info)
-    _dump_response_data(response, prefixes, data)
+                       proxy_info=proxy_info, sanitizer=sanitizer)
+    _dump_response_data(response, prefixes, data, sanitizer=sanitizer)
     return data
 
 
-def dump_all(response, request_prefix=b'< ', response_prefix=b'> '):
+def dump_all(response, request_prefix=b'< ', response_prefix=b'> ',
+             sanitizer=NOOP_SANITIZER):
```
sigmavirus24: Same comment as above.

tests/test_dump.py
```diff
@@ -18,6 +18,13 @@
 
 from . import get_betamax
 
+# A semi-random selection of "normal" headers not touched by sanitizing
+NORMAL_HEADERS = (
+    "Accept", "Accept-Encoding", "Host", "User-Agent", "Accept-Ranges",
+    "Cache-Control", "Content-Encoding", "Content-Length", "Content-Type",
+    "Date", "Etag", "Expires", "Last-Modified", "Server", "Vary", "X-Cache"
```
sigmavirus24: What about custom headers? `OpenStack-Identity`, `X-My-Custom-Header`, etc.?

requests_toolbelt/utils/dump.py
```diff
@@ -3,19 +3,116 @@
...
 
 
+class Sanitizer(object):
+    """Performs no sanitation"""
+
+    def request_header(self, name, value):
+        """Sanitize a request header
+
+        :param name: The header name
+        :type name: `compat.basestring`
```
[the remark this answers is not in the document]
mortenlj: It references the type `compat.basestring`, which is already in use in this module. However Sphinx doesn't know it, and since I'm using `bytes` elsewhere I guess the convention is to use Python 3 types. I'll change it to `str`.

requests_toolbelt/utils/dump.py
mortenlj: My suggestion was to provide a method for providing your own, and one implementation that will do the right thing for most people. I imagine most people who need this will need to support this exact use-case: Redact headers commonly used to send sensitive data.

To me, the value of the feature increases tenfold if this implementation is provided, instead of every user of toolbelt having to implement this on their own in every project where this is needed.

Of course, this is your library, if you don't want it I can remove it.

requests_toolbelt/utils/dump.py
mortenlj: Oops, will fix.

requests_toolbelt/utils/dump.py
mortenlj: I did it like this because iterating over the headers and values is non-trivial, and I thought it would be more useful if you didn't have to figure out how to do that every time you want to implement a sanitizer. Also, since headers is not a dict, what should this method return in order to replace it? Or should it modify the headers object in-place (is that possible)?

With regards to stripping a name-value pair completely, I figured that was a very uncommon use-case that might not be worth supporting.

I can change it to work on the full headers, but then I need help understanding the questions above.

requests_toolbelt/utils/dump.py
mortenlj: I'm not sure I understand how this is unsafe behavior?

Whenever someone needs to implement a sanitizer, they would need to subclass the `NoopSanitizer` instead of the "interface", because it's going to be extremely rare that you want to override all four methods. I'm not sure I see the benefit of splitting the default behavior out from the interface class, but I can do it if you think it is important.

requests_toolbelt/utils/dump.py
mortenlj: Why? I don't see the upside to be honest, so it would be appreciated if you could enlighten me.

Downsides of doing as you suggest (as I see it):
- Hide the actual behavior from the signature, since you no longer see that it uses `NoopSanitizer`
- Complicate (slightly) the body of the function, since we now have to do a `None`-check and instantiate a new `NoopSanitizer`
- Performance-drop if we instantiate a `NoopSantiizer` on every call

tests/test_dump.py
mortenlj: This is used in a test to confirm that we are not redacting all headers. A test that covers all possible custom headers would be impossible, since that is a nearly infinite set. I decided to just use a semi-random selection of common headers.

If we are removing the `HeaderSanitizer`, then this point is moot anyway, since this test would go away too.

requests_toolbelt/utils/dump.py
sigmavirus24: > Also, since headers is not a dict, what should this method return in order to replace it? 

So headers is going to be a requests-specific dictionary implementation: `CaseInsensitiveDict`. We can, do this:

```
sanitized_headers = headers.copy()
for name in headers:
    if self.should_sanitize_header(name):
        sanitized_headers[name] = REDACTED_VALUE
    elif self.should_strip_header(name):
        del sanitized_headers[name]
```

This would then allow users to define `should_sanitize_header` and `should_strip_header` which only needs to operate on the header name and return True/False. This could be the default implementation for the `_header` sanitization functions since calling the `should_*_header` would raise exceptions if not implemented by default.

requests_toolbelt/utils/dump.py
sigmavirus24: My concern is that this library is basically on life-support. Meaning that as this list grows, who will maintain it? And if people think this list is comprehensive, they will likely leak sensitive data without meaning to because they think this library is maintaining the list themselves.

Alternatively, what if we do the following:

1. Update `HeaderSanitizer` to require the headers set  be provided by the user.
1. Document the `SENSITIVE_HEADERS` list more thoroughly as a point-in-time snapshot from the repository with a Last-Updated date of today or whatever.
1. Indicate that if there are updates to the list that weren't made here users should use that themselves and/or send a PR to update this list.

requests_toolbelt/utils/dump.py
sigmavirus24: > Hide the actual behavior from the signature, since you no longer see that it uses NoopSanitizer

The behaviour should be documented in the doc-string regardless. I'm not sure the behaviour belongs in the signature at all.

> Complicate (slightly) the body of the function, since we now have to do a None-check and instantiate a new NoopSanitizer

I don't believe that `sanitizer = sanitizer or NOOP_SANITIZER` or even `sanitizer = sanitizer or NoopSanitizer()` is that big of a complication.

> Performance-drop if we instantiate a NoopSantiizer on every call

```
❯❯❯ python -m timeit -s 'from requests_toolbelt.utils.dump import Sanitizer' -- 'Sanitizer()'    ⏎ feature/redact-sensitive-headers
10000000 loops, best of 3: 0.108 usec per loop
```

And that's on a sad little old machine. Anything remotely modern or performant should have little trouble with the added instantiation.

Further, comparing it to the rest of the `dump_*` functions which iterate over potentially large dictionaries, it really can't be such a significant hit as to prefer the minor potential performance improvement over correctness.

requests_toolbelt/utils/dump.py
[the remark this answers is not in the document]
mortenlj: Strictly speaking I'm not 100% sure what type is expected here. It depends on what types are used in the headers. I believe it is `unicode` in Python 2 and `str` in Python 3, but I'm not sure.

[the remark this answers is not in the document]
sigmavirus24: Generally speaking, it's roughly `Union[six.bytes_type, six.text_type]` where on Python 2 those are `str` and `unicode` respectively and on Python 3 they are `bytes` and `str`.

requests_toolbelt/utils/dump.py
mortenlj: I'm confused. Above you talk about not wanting a list of headers to sanitize by default, yet here you are talking about making the default do sanitizing because that is safer. If you don't have a list of headers to sanitize, how do you intend to do "safe-by-default" sanitizing?

If we say that the default is to not do anything (as today, aka unsafe-by-default), then a `HeaderSanitizer` that requires the user to provide a list of headers to sanitize makes sense. The user would the create an instance with a list of headers that they consider sensitive, and pass that as the `sanitizer` argument.

requests_toolbelt/utils/dump.py
mortenlj: Ok, I think I get the gist of your suggestion. I will try to make some changes to move in this direction, and we can see where that gets us.

requests_toolbelt/utils/dump.py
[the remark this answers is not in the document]
mortenlj: So `requests.compat.basestring` is probably the most correct thing to use then?

requests_toolbelt/utils/dump.py
mortenlj: I still don't see the upside to be honest, but as you say, most of my objections are really minor points. I'll change it.

mortenlj:
I have made some changes, that I believe are in the right direction. I think there are some comments I have yet to address, but I'm not sure if they are still relevant. I think the discussion might have changed the direction a bit?

Anyway, let me know what you think about these latest changes, and where you want further changes (if any). I want to get this done, but need you to tell me where to go so I don't have to guess.

requests_toolbelt/utils/dump.py
sigmavirus24: Hi there, let me be clear:

If the base sanitizer never sanitizes and silently is a NoOp then it's a bad default. The base sanitizer that users should sub-class should provide errors when trying to use the other methods because the user hasn't defined what they want to do. Otherwise `Sanitizer` in this case is unsafe-by-default. Safe-by-default means making the user think about it in this case.

Further, I'm starting to think this design is too limiting. How can someone say, compose the HeaderSanitizer here and a BodySanitizer elsewhere. They'd have to sub-class and then that becomes a nightmare of copy-pasting code around to make it all play nicely together. Instead, we likely want to accept a list of sanitizers that the user can specify, and call them in the order the user provides.

jdufresne:
Hi @mortenlj and @sigmavirus24 this is a feature I'm really interested in. Is there anything I can do to help move this along?

mortenlj:
@jdufresne: I've moved on to other things, probably not going to circle back to this for the foreseeable future. Feel free to take over.

PR #231, opened 2018-10-28
Add an SSLContextAdapter

nikosgraser:
An adapter that allows more fine-grained control over the encryption
parameters used in `requests` by passing an `SSLContext` object
directly.
Originally borne from the desire to test a different implementation of
Python's `ssl` module's `create_default_context()` with `requests`.

### DIFF
--- a/requests_toolbelt/adapters/ssl_context.py
+++ b/requests_toolbelt/adapters/ssl_context.py
@@ -1,0 +1,51 @@
+# -*- coding: utf-8 -*-
+"""This file contains an implementation of the SSLContextAdapter.
+
+   It requires a version of requests >= 2.4.0.
+"""
+
+from requests.adapters import HTTPAdapter
+
+
+class SSLContextAdapter(HTTPAdapter):
+    """An adapter that lets the user inject a custom SSL context for all
+       requests made through it.
+
+       The SSL context will simply be passed through to urllib3, which
+       causes it to skip creation of its own context.
+
+       Note that the SSLContext is not persisted when pickling - this is on
+       purpose.
+       So, after unpickling the SSLContextAdapter will behave like an
+       HTTPAdapter until a new SSLContext is set.
+
+       Example usage:
+
+       .. code-block:: python
+
+           import requests
+           from ssl import create_default_context
+           from requests import Session
+           from requests_toolbelt.adapters.ssl_context import SSLContextAdapter
+
+           s = Session()
+           s.mount('https://', SSLContextAdapter(ssl_context=create_default_context()))
+    """
+
+    def __init__(self, **kwargs):
+        self.ssl_context = None
+        if 'ssl_context' in kwargs:
+            self.ssl_context = kwargs['ssl_context']
+            del kwargs['ssl_context']
+
+        super(SSLContextAdapter, self).__init__(**kwargs)
+
+    def __setstate__(self, state):
+        # SSLContext objects aren't picklable and shouldn't be persisted anyway
+        self.ssl_context = None
+        super(SSLContextAdapter, self).__setstate__(state)
+
+    def init_poolmanager(self, *args, **kwargs):
+        if 'ssl_context' not in kwargs:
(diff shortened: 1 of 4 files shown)


requests_toolbelt/adapters/ssl_context.py
```diff
@@ -0,0 +1,53 @@
...
+
+   It requires a version of requests >= 2.4.0.
+"""
+
+import requests
+from requests.adapters import HTTPAdapter
+
+
+class SSLContextAdapter(HTTPAdapter):
+    """An adapter that lets the user inject a custom SSL context for all
```
sigmavirus24: Please follow the style of other class and module docstrings in this library.

requests_toolbelt/adapters/ssl_context.py
```diff
@@ -0,0 +1,53 @@
+# -*- coding: utf-8 -*-
+"""This file contains an implementation of the SSLContextAdapter.
+
+   It requires a version of requests >= 2.4.0.
```
sigmavirus24: Please look at other submodules for the library's style and follow that

requests_toolbelt/adapters/ssl_context.py
```diff
@@ -0,0 +1,53 @@
...
+            del kwargs['ssl_context']
+
+        super(SSLContextAdapter, self).__init__(**kwargs)
+
+    def __setstate__(self, state):
+        # SSLContext objects aren't picklable and shouldn't be persisted anyway
+        self.ssl_context = None
+        super(SSLContextAdapter, self).__setstate__(state)
+
+    def init_poolmanager(self, *args, **kwargs):
```
sigmavirus24: Please add docstrings to these methods.

tests/test_ssl_context_adapter.py
```diff
@@ -0,0 +1,60 @@
...
+import pytest
+import requests
+from requests.adapters import HTTPAdapter
+from requests_toolbelt.adapters.ssl_context import SSLContextAdapter
+
+
+@pytest.mark.skipif(requests.__build__ < 0x020400,
+                    reason="Test case for newer requests versions.")
+@mock.patch.object(HTTPAdapter, 'init_poolmanager')
+def test_ssl_context_arg_is_passed_on_newer_requests(init_poolmanager):
```
sigmavirus24: Please add docstrings to the test functions

tests/test_ssl_context_adapter.py
```diff
@@ -0,0 +1,60 @@
...
+    )
+    init_poolmanager.assert_called_once_with(
+        10, 5, block=True, ssl_context=ssl_context
+    )
+
+
+@pytest.mark.skipif(requests.__build__ >= 0x020400,
+                    reason="Test case for older requests versions.")
+@mock.patch.object(HTTPAdapter, 'init_poolmanager')
+def test_ssl_context_arg_is_not_passed_on_older_requests(init_poolmanager):
```
sigmavirus24: Here too

requests_toolbelt/adapters/ssl_context.py
nikosgraser: Hi @sigmavirus24, 

I'm not sure what kind of information would be expected here in a docstring.

The only useful information I could come up with would be something like "Calls superclass' init_poolmanager method with an extra arg on requests >= 2.4.0". This seems redundant, given that the methods themselves are really not complicated.

Adding docstrings here would also be inconsistent with the implementations of the other Adapters in this module. 

If all methods are supposed to have some kind of docstring despite these arguments, I'll be happy to provide some - just trying to understand the reasoning here.

Regards
Nikos

tests/test_ssl_context_adapter.py
nikosgraser: See also my comment on docstrings in the actual module - I've added slightly wordier descriptions as docstrings now. Is that what you expected @sigmavirus24?

nikosgraser:
Ping @sigmavirus24 could you have another look? Thanks :)

PR #234, opened 2018-12-13
Add pkcs12 support

rashley-iqt:
An adapter to be used in cases where a user needs to authenticate to a service using a .p12/.pfx certificate and it is not desirable or feasible to first convert that certificate into a .pem format.  Born out of a use case in which the existence of temporary files containing cert data were deemed too great a risk.


sigmavirus24:
I think this looks okay to me (There are some style nits but those aren't as important to me in the context of this PR). I'm asking someone else with a bit more experience around OpenSSL and Pkcs12 to review this to make sure I'm not missing something crucial.

alex:
Hey, providing some feedback at @sigmavirus24's request.

1. I don't think it makes sense to organize this as an `HttpAdapter`, that'd mean you need one for every serialization format. Organizing this to just deserialize the data and then pass that to an `HttpAdapter` that can take keys/certs as bytes (can the primary one do this?)

2. I wouldn't use the pyOpenSSL PKCS#12 implementation. pyca/cryptography's next release includes PKCS#12 support, so I'd wait for that https://cryptography.io/en/latest/hazmat/primitives/asymmetric/serialization/#pkcs12 instead of using pyOpenSSL.

3. FWIW, stdlib SSL (and I assume therefore requests) supports password protected private keys, so it shouldn't be necessary to use PKCS#12 if that's all you need.

rashley-iqt:
@alex thanks for the feedback, I really appreciate it.  
1.  My understanding of the primary `HttpAdapter`, and somebody please correct me if I am wrong about this, is that it will accept a string path to a certificate file that must be in .pem format. I implemented is as an `HttpAdapter` because based on [this comment](https://github.com/requests/requests/issues/1573#issuecomment-353686689) and [this comment](https://github.com/requests/requests/issues/2118#issuecomment-47803629) it seemed to be the preferred method for handling this type of scenario. I agree that it wouldn't be a good idea to need a separate `HttpAdapter` for every potential serialization format and I would actually like to grow this adapter out to support other formats, such as PKCS#11.  If there is interest on the part of @sigmavirus24 and the other maintainers, would it be acceptable to treat this as a starting point and add other serialization methods? If so, would it make sense to change the names of files and classes to make that intention clearer?
1. I would actually prefer to use pyca/cryptography but I don't know when a version that will support my needs will be released.  Once that version is available I am willing to commit to refactoring my code to use it instead.  Do you have any insight into when that might be?
1. For my current project .p12 files are a strict requirement and decomposing them into temporary .pem files is not an option for us, hence the need for this adapter.

rashley-iqt:
@alex you should be more like this? 
```
import requests
from OpenSSL.crypto import load_pkcs12
from requests_toolbelt.adapters import X509Adapter

p12 = load_pkcs12(pkcs12_data, pkcs12_password_bytes)
cert = p12.get_certificate()
private_key = p12.get_privatekey()

s = requests.Session()
a = X509Adapter(max_retries=3, cert_bytes=cert, private_key_bytes=private_key)
s.mount('https://', a)
```

alex:
Something like that, yeah! And maybe it should even be on the default `requests.Session` or something.

rashley-iqt:
I would agree, but that is a question for @sigmavirus24 because then we would need to move from the toolbelt to requests/requests since that is where `HttpAdapter` lives.

sigmavirus24:
I'd be happy to have that in the toolbelt today. And work towards getting it exempted from requests' feature-freeze for sometime in the future if we can show there's a lot of usage of it.

rashley-iqt:
@sigmavirus24 i've got the changes you requested mostly implemented, but I am currently struggling with betamax.  When i perform my tests locally and can access the server i am using for testing the tests pass.  However, when I disconnect the network and rely on the cassette file the tests fail.  I notice that the `http_interactions` property is being populated with an empty array which makes me think that I have done something incorrect in terms of how I have the recorder configured but I am not sure what.  Do you have any thoughts or advice?  Is there a log that is populated somewhere?

requests_toolbelt/adapters/x509_adapter.py
```diff
@@ -0,0 +1,179 @@
...
+
+from datetime import datetime
+from requests.adapters import HTTPAdapter
+import requests
+
+from .._compat import PyOpenSSLContext
+from .. import exceptions as exc
+
+try:
+    from _ssl import PROTOCOL_TLS as PROTOCOL
```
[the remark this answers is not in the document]
rashley-iqt: i used `_ssl` because i didn't need any of the actual implementation portions of ssl, just some of the constants that are defined in `_ssl` and passed through `ssl`.  It also solved an issue i was having trying to import directly from `ssl` that only occurred in the tox py27 environment.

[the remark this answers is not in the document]
sigmavirus24: I'm surprised that you ran into issues with importing `ssl` on Py27. Would you be willing to add a comment explaining that? I just don't want to come back to this in the future, change the import and break something for someone.

What version of Python 2.7 are you using by the way?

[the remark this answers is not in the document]
rashley-iqt: I'll add a comment explaining that.  I'm using 2.7.15, but it actually happened under tox in 2.7.14.  I tried several things to try to correct it, but the switch to using _ssl was the only thing that worked on the 2.7.x and 3.x lines.

requests_toolbelt/adapters/x509.py
```diff
@@ -0,0 +1,184 @@
...
+        failed connections. If you need granular control over the
+        conditions under which we retry a request, import urllib3's
+        ``Retry`` class and pass that instead.
+    :param pool_block: Whether the connection pool should block for
+            connections.
+
+    :param \**kwargs:
+        see below
+
+    :Keyword Arguments:
```
sigmavirus24: It seems my request about this documentation being corrected was lost.

requests_toolbelt/adapters/x509.py
```diff
@@ -0,0 +1,184 @@
...
+                "installed. Version {0} was found instead.".format(
+                    requests.__version__
+                )
+            )
+
+
+def check_cert_dates(cert):
+    """Verify that the supplied client cert is not invalid."""
+
+    if (
```
sigmavirus24: It also seems my request to import the datetime module and find `utcnow()` was also lost.

requests_toolbelt/adapters/x509_adapter.py
```diff
@@ -0,0 +1,173 @@
...
+        socket connections and connection timeouts, never to requests where
+        data has made it to the server. By default, Requests does not retry
+        failed connections. If you need granular control over the
+        conditions under which we retry a request, import urllib3's
+        ``Retry`` class and pass that instead.
+    :param pool_block: Whether the connection pool should block for
+            connections.
+
+    :param bytes cert_bytes:
+        bytes object containing contents of a cryptography.x509Certificate
```
sigmavirus24: I see what happened there are two files here, the `x509_adapter` and `x509` are not in sync properly and one has the changes I wanted while the other has the module name i wanted. Can you reconcile these please?

rashley-iqt: sorry, something went weird with the rename and merge.  I've corrected this.

sigmavirus24: No need to apologize :) Thanks for fixing it up

sigmavirus24:
Alex gave this an informal :+1: on IRC, so I'm going to merge it. Thanks @rashley-iqt for following through and tackling this.

rashley-iqt:
@sigmavirus24 do you have any idea when the next release of the toolbelt is scheduled for?

sigmavirus24:
There isn't a schedule. We need to put together release notes though before we can release. If you want to send a PR with those notes then that can help accelerate this as the weather where I am should keep me outside shoveling snow probably through the end of my life 😜

ryanwilsonperkin:
Hey folks, I suspect this might have introduced a breaking change into this project. We're installing this package as a dependency of `twine` (just a simple `pip install twine`) and now finding that we're getting an error as a result of trying to import `pyopenssl`.

Full stack trace:
```python
  File "/usr/local/bin/twine", line 6, in <module>
    from twine.__main__ import main
  File "/usr/local/lib/python3.6/site-packages/twine/__main__.py", line 23, in <module>
    from twine.cli import dispatch
  File "/usr/local/lib/python3.6/site-packages/twine/cli.py", line 23, in <module>
    import requests_toolbelt
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/__init__.py", line 12, in <module>
    from .adapters import SSLAdapter, SourceAddressAdapter
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/adapters/__init__.py", line 12, in <module>
    from .ssl import SSLAdapter
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/adapters/ssl.py", line 16, in <module>
    from .._compat import poolmanager
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/_compat.py", line 59, in <module>
    from urllib3.contrib.pyopenssl import PyOpenSSLContext
  File "/usr/local/lib/python3.6/site-packages/urllib3/contrib/pyopenssl.py", line 46, in <module>
    import OpenSSL.SSL
ModuleNotFoundError: No module named 'OpenSSL'
```

PR #240, opened 2019-01-29
Update history to reflect commits since 0.8.0

rashley-iqt:
This is an initial draft of release notes, as requested.  Please let me know what it needs.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -1,6 +1,42 @@
 History
 =======
 
+0.8.1 -- 2019-01-28
+-------------------
+
+More information about this release can be found on the `0.8.1 milestone`_.
+
+New Features
+~~~~~~~~~~~~
+
+- Add X509 Adapter that can handle PKCS12 
+- Add stateless solution for streaming files by MultipartEncoder from one host to another (in chunks)
+
+Fixed Bugs
+~~~~~~~~~~
+
+- Update link to example
+-  Move import of ``ABCs`` from collections into version-specific part of _compat module
+- Fix backwards incompatibility in ``get_encodings_from_content``
+- Correct callback documentation for ``MultipartEncoderMonitor``
+- Fix bug when ``MultipartEncoder`` is asked to encode zero parts
+- Correct the type of non string request body dumps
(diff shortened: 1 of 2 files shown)


HISTORY.rst
```diff
@@ -1,6 +1,42 @@
...
+- Corrected parentheses match in example usage
+- Fix "oject" to "object" in ``MultipartEncoder``
+- Fix URL for the project after the move 
+- Add fix for OSX TCPKeepAliveAdapter
+
+Miscellaneous
+~~~~~~~~~~~~~
+- Remove py33 from testing and add Python 3.6 and nightly testing to the travis matrix.
+
+.. _0.8.1 milestone:
```
sigmavirus24: This link needs to resolve somewhere. If there's no milestone, we can drop the reference here and in the introduction.

HISTORY.rst
```diff
@@ -1,6 +1,42 @@
...
+- Coerce proxy URL to bytes before dumping request
+- Avoid bailing out with exception upon empty response reason
+- Corrected Pool documentation
+- Corrected parentheses match in example usage
+- Fix "oject" to "object" in ``MultipartEncoder``
+- Fix URL for the project after the move 
+- Add fix for OSX TCPKeepAliveAdapter
+
+Miscellaneous
+~~~~~~~~~~~~~
```
sigmavirus24: Can you add a new line between the `~` and the list item?

HISTORY.rst
```diff
@@ -1,6 +1,42 @@
...
 
+0.8.1 -- 2019-01-28
+-------------------
+
+More information about this release can be found on the `0.8.1 milestone`_.
+
+New Features
+~~~~~~~~~~~~
+
+- Add X509 Adapter that can handle PKCS12
```
sigmavirus24: Since there's new features the release should be `0.9.0`.

Also if you can make the date the 29th or the 30th, that'd be perfect. (I like the date in the history to match the date that I actually release it).

sigmavirus24: Also we should bump the version of the actual package (in `requests_toolbelt/__init__.py` or there-abouts) to match the version here.

Issue #241, opened 2019-01-29
Missing dependency/crash with 0.9.0


piyushk:
I use twine to release packages to an internal repo. Our last CI build downloaded requests-toolbelt 0.9.0 and crashed with the following error:
```
Traceback (most recent call last):
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/_compat.py", line 56, in <module>
    from requests.packages.urllib3.contrib.pyopenssl \
  File "/usr/local/lib/python3.6/site-packages/urllib3/contrib/pyopenssl.py", line 46, in <module>
    import OpenSSL.SSL
ModuleNotFoundError: No module named 'OpenSSL'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/usr/local/lib/python3.6/runpy.py", line 193, in _run_module_as_main
    "__main__", mod_spec)
  File "/usr/local/lib/python3.6/runpy.py", line 85, in _run_code
    exec(code, run_globals)
  File "/usr/local/lib/python3.6/site-packages/twine/__main__.py", line 20, in <module>
    from twine.cli import dispatch
  File "/usr/local/lib/python3.6/site-packages/twine/cli.py", line 23, in <module>
    import requests_toolbelt
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/__init__.py", line 12, in <module>
    from .adapters import SSLAdapter, SourceAddressAdapter
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/adapters/__init__.py", line 12, in <module>
    from .ssl import SSLAdapter
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/adapters/ssl.py", line 16, in <module>
    from .._compat import poolmanager
  File "/usr/local/lib/python3.6/site-packages/requests_toolbelt/_compat.py", line 59, in <module>
    from urllib3.contrib.pyopenssl import PyOpenSSLContext
  File "/usr/local/lib/python3.6/site-packages/urllib3/contrib/pyopenssl.py", line 46, in <module>
    import OpenSSL.SSL
ModuleNotFoundError: No module named 'OpenSSL'
```

It seems pyopenssl was not installed as a dependency for `requests-toolbelt==0.9.0`. The following packages where installed when installing `twine==1.11.0`:
```
Collecting twine==1.11.0
[...]

PR #242, opened 2019-01-29
Add pyopenssl to install_requires

altendky:
https://github.com/requests/toolbelt/commit/675fab790b0e9fefd6beb0f38d9a5da4b056c59c
https://github.com/requests/toolbelt/issues/241

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
     packages=packages,
     package_data={'': ['LICENSE', 'AUTHORS.rst']},
     include_package_data=True,
-    install_requires=['requests>=2.0.1,<3.0.0'],
+    install_requires=['requests>=2.0.1,<3.0.0', 'pyopenssl'],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'License :: OSI Approved :: Apache Software License',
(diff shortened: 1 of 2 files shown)


jeid64:
https://github.com/requests/toolbelt/commit/675fab790b0e9fefd6beb0f38d9a5da4b056c59c
Wasn't caught in testing because the tox configuration installs it.

altendky:
@jeid64 good point.  Let's see if it works without those.

sigmavirus24:
Can we instead just try except around the import that's offending and ignore it. I think the adapter should be resilient to not having the pyopenssl backend. That would be better than blanket requiring pyopenssl here.

ryanwilsonperkin:
@sigmavirus24 there does appear to be a try/except around it, but not working properly. Perhaps the condition that's checking the requests version is incorrect?

https://github.com/requests/toolbelt/blob/da420bdfd9f8baf2041ee2cf1a666e1fc59b1551/requests_toolbelt/_compat.py#L52-L59

sigmavirus24:
No, I believe the unconditional import from `*urllib3.contrib.pyopenssl` whether vendored or not is the problem here and I missed that during review.

ryanwilsonperkin:
@sigmavirus24 gotcha. How about this? https://github.com/requests/toolbelt/pull/243

sigmavirus24:
Superseded by #243

sigmavirus24:
Thanks for sending this @altendky

PR #243, opened 2019-01-29
Handle missing pyopenssl dependency

ryanwilsonperkin:
Alternative to #242
Fixes #241 

Handle the case where `pyopenssl` is not installed and fall back to default of setting `PyOpenSSLContext` to `None`.

PR #244, opened 2019-01-30
Add tox test

greatestape:
This adds:

- another tox environment that explicitly excludes `pyopenssl`.
- a test that the `_compat` package doesn't blow up when openssl is missing
- skip decorators on a couple other tests that do need openssl to run

Follow up to #243. Adds coverage to guard against issue that caused #241

### DIFF
--- a/tests/test_compat.py
+++ b/tests/test_compat.py
@@ -3,11 +3,11 @@ import pytest
 try:
     import OpenSSL
 except ImportError:
-    OPENSSL_AVAILABLE = False
+    PYOPENSSL_AVAILABLE = False
 else:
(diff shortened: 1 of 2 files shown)


tests/test_compat.py
```diff
@@ -0,0 +1,14 @@
+import pytest
+
+try:
+    import OpenSSL
+except ImportError:
+    OPENSSL_AVAILABLE = False
```
sigmavirus24: Definitely a nit, and I'd understand if you don't want to update it, but can this be `pyOpenSSL` since that's what is actually missing.

greatestape: Fixed

PR #290, opened 2020-04-02
Fixes code highlighting in README.rst

sobolevn:


README.rst
```diff
@@ -56,6 +56,8 @@ User-Agent constructor
 
 You can easily construct a requests-style ``User-Agent`` string::
```
sigmavirus24, line 57: This line needs to change too if this change were to be accepted

PR #300, opened 2020-11-02
Handle streamed content

neg3ntropy:


requests_toolbelt/utils/dump.py
```diff
@@ -104,7 +104,10 @@ def _dump_response_data(response, prefixes, bytearr):
 
     bytearr.extend(prefix + b'\r\n')
 
-    bytearr.extend(response.content)
+    if not response._content_consumed or response._content is False:
```
sigmavirus24, line 107: I think we avoid private attributes everywhere else in the code base. I wonder if something a little simpler is easier:

```py
>>> r = requests.get("https://httpbin.org/get")  
>>> r.raw.closed
True
>>> s = requests.get('https://httpbin.org/get', stream=True)
>>> s.raw.closed
False
```

neg3ntropy: Good catch, that seems to be working just as well. I updated the PR and added a test

neg3ntropy:
I don't think the failing tests have anything to do with what I changed

PR #302, opened 2021-01-11
Fixing missing newline in dump

noxora:
Hi there! 
It would make the dump clearer and follow the convention of the rest of the dump to include these newlines, would we be able to do so?

### DIFF
--- a/requests_toolbelt/utils/dump.py
+++ b/requests_toolbelt/utils/dump.py
@@ -81,6 +81,7 @@ def _dump_request_data(request, prefixes, bytearr, proxy_info=None):
             # In the event that the body is a file-like object, let's not try
             # to read everything into memory.
             bytearr.extend(b'<< Request body is not a string-like type >>')
+        bytearr.extend(b'\r\n')
     bytearr.extend(b'\r\n')
 
 


requests_toolbelt/utils/dump.py
```diff
@@ -76,11 +76,11 @@ def _dump_request_data(request, prefixes, bytearr, proxy_info=None):
...
     if request.body:
         if isinstance(request.body, compat.basestring):
-            bytearr.extend(prefix + _coerce_to_bytes(request.body))
+            bytearr.extend(prefix + _coerce_to_bytes(request.body) + b'\r\n')
         else:
             # In the event that the body is a file-like object, let's not try
             # to read everything into memory.
-            bytearr.extend(b'<< Request body is not a string-like type >>')
+            bytearr.extend(b'<< Request body is not a string-like type >>' + b'\r\n')
     bytearr.extend(b'\r\n')
```
sigmavirus24, line 84: I feel like that's what this line was intended to do. Can you give me an example of before/after your change here?

requests_toolbelt/utils/dump.py
```diff
@@ -76,11 +76,11 @@ def _dump_request_data(request, prefixes, bytearr, proxy_info=None):
     bytearr.extend(prefix + b'\r\n')
     if request.body:
         if isinstance(request.body, compat.basestring):
-            bytearr.extend(prefix + _coerce_to_bytes(request.body))
+            bytearr.extend(prefix + _coerce_to_bytes(request.body) + b'\r\n')
         else:
             # In the event that the body is a file-like object, let's not try
             # to read everything into memory.
-            bytearr.extend(b'<< Request body is not a string-like type >>')
+            bytearr.extend(b'<< Request body is not a string-like type >>' + b'\r\n')
```
sigmavirus24, line 83: We don't need string concatenation here, just add `\r\n` after `>>` and before `'`

noxora: You're right. I'll fix this

requests_toolbelt/utils/dump.py
noxora: I'll get this to you asap

noxora: This is before
image

This is after 
image

The code which generated this was 
```import requests
from requests_toolbelt.utils import dump


def main():
    response = requests.get(
        "http://httpbin.org",
        data="This is some test data that appears directly above the response's status code",
    )
    print(
        dump.dump_response(response, request_prefix=b"", response_prefix=b"").decode()
    )
    # print(response)


main()```

sigmavirus24: Actually, instead of updating those two lines, why not add an extra `bytes.append(b'\r\n')` inside the `if request.body` block

noxora: Would that be after the `if isinstance ... else` block, but still inside the `if request.body` block? I'll just squash the commits together once we're happy with it

sigmavirus24: Yes that's where I meant I'd like it. Once you squash the commits we can merge this I think

noxora: Squashed. Merge whenever, thanks for your help! 😊

PR #318, opened 2021-10-21
Fix implementation error

ghost:
Error details:
Whenever we call the utils.dump.dump_response with the parameters, we expect it to dump the response only.

In the current implementation(before this fix), it dumped both the request and the response, we dont want that.

This "error" has now been fixed with this commit by removing the line that dumps the request in the corresponding function.


V0ldemrt:
Also, we should probably removed the "request_prefix" option in the function because we dont require it anymore.

I forgot to remove it in this PR

sigmavirus24:
The docstring clearly defines that it's dumping the "request-response cycle" not "just the response"

V0ldemrt:
But then how is it different from lets say dump_app

PR #323, opened 2022-01-20
Update _compat.py

zjzh:
refactoring code with var unpack which is more pythonic, concise, readable and efficient; how do think this change which has practical value？


sigmavirus24 (commented):
This library still nominally supports Python 2 and thus can't accept this

PR #326, opened 2022-01-24
Strip empty fields in MultipartEncoder

rdbisme:
This is coherent with [what's done in requests](https://github.com/psf/requests/blob/main/requests/models.py#L163). My usecase concerns using [responses.matchers.multipart_matcher], which uses bare requests to encode the multipart request, while the client, that specifically is [Gitlab's` one](https://github.com/python-gitlab/python-gitlab/blob/30117a3b6a8ee24362de798b2fa596a343b8774f/gitlab/client.py#L599), uses `MultipartEncoder`. The two bodies for the request are different.


rdbisme:
Up? :)

PR #333, opened 2022-10-11
Run twine check in CI

pquentin:
Otherwise, errors in HISTORY.rst can prevent uploads to PyPI. This is what happened with the last release which had a syntax error in HISTORY.rst. So 0.10.0 on PyPI isn't actually what I tagged in Git, since I had to add that dash in HISTORY.rst to upload it. I won't do it again thanks to this pull request. This code is based on urllib3 GitHub Actions.

Example failure: https://github.com/pquentin/toolbelt/actions/runs/3227044314/jobs/5281320476 
Example success: https://github.com/pquentin/toolbelt/actions/runs/3227081530

@achapkowski Can you please take a look?


achapkowski:
Look good, but the success has a warning: WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`.   
This need to be addressed by adding long_description_content_type in the setup.py and set the value to "text/s-rst".

Should we also add a linter check and enforce psf/black standards?

achapkowski (changes requested):
See comment in PR about adding to setup.py and optional addition of psf/black linter.

pquentin:
Thank you @achapkowski, I fixed the warnings and made sure we won't have any more of them with `--strict`. Result: https://github.com/pquentin/toolbelt/actions/runs/3241160689/jobs/5312723914

While I use black, isort and various linters in my projects, I'd like to avoid making big changes here for now. My goal is still to confidently release a version without urllib3 warnings. So my next step is running tests in GitHub Actions, marking some of them as xfail. I'm moving slowly as I'm on vacation. But maybe you want to work on that?

achapkowski (approved):
Approved, thank you for updating the setup.py

achapkowski:
@pquentin I can take a look at the warnings and correct them as they pop-up.

PR #334, opened 2022-10-17
Run tests using GitHub Actions

pquentin:
I think it's important to have a minimal green CI before making more substantial changes. The current limitations:
 * We're not running against multiple versions of requests - we'll have to decide how helpful it is anyway, given 2.1.0 and 2.2.1 are old now
 * We're not running the docs, docstring and readme jobs. (I tried docstrings but it failed -- just realized now that the failure was allowed in Travis!)
 * We're running tests on Python 2.7. I just checked and only [1% of users still use Python 2.7](https://pypistats.org/packages/requests-toolbelt) so it's safe to drop but I think it should be done explicitly
 *  Various tests don't work so I marked them as xfail. We'll need to fix the X509 tests before fixing the warnings

@achapkowski Can you please take a look?


sigmavirus24:
LGTM

PR #335, opened 2022-10-24
Fix x509 tests by using trustme

pquentin:
This way the certificate won't be able to expire anymore and we won't have to store a blob in git. trustme only supports PEM, not DER, which is why test_x509_der has to re-export the certificate and key to DER.

Additionally, since trustme does not support password-protected keys, the re-export was the perfect place to add a password, so test_x509_der also tests the password case, while it was test_x509_pem until now.

The tests are still not end-to-end, we're just running the x509.py code without actually establishing a TLS connection. This could be fixed at a later point, but is considered out of scope here as that would be a new feature.

@achapkowski @sethmlarson @sigmavirus24  Can you please take a look?


achapkowski:
Since we are fixing tests, can you drop the Python 2.7 tests and actions?

bmwiedemann:
Worth noting that current trustme stops working in 2038: https://github.com/python-trio/trustme/pull/42

PR #336, opened 2022-10-24
Fix urllib3 warning

### LINKED ISSUE #331: Remove import from deprecated urllib3.contrib.pyopenssl
urllib3 is in the process of deprecating urllib3.contrib.pyopenssl, see https://github.com/urllib3/urllib3/issues/2680

requests-toolbelt currently attempts to import `from urllib3.contrib.pyopenssl import PyOpenSSLContext`. Apparently this is used by `X509Adapter`.

What's the path forward here? Can `X509Adapter` be ported to newer APIs that are supported?

Otherwise it would be preferred if DeprecationWarning was only shown for code that actually uses `X509Adapter`, maybe make the import lazy and put it in a function?

pquentin:
This will fix the warning for urllib3 1.26.x. To support urllib3 2.x we will need to vendor PyOpenSSLContext instead.

This builds on top of #335 but can also be merged directly if we prefer.

Resolves #331, Closes #335

@achapkowski @sethmlarson @sigmavirus24 Can you please take a look?


achapkowski (approved):
approved

PR #337, opened 2022-10-24
removes older versions of python. 

achapkowski:
removes earlier versions of python. Drops support for Python 2.7 and Python <3.7


achapkowski:
@pquentin I went ahead and dropped older versions of Python and fixed the tests.

pquentin (changes requested):
Thanks! Can you please:

 * Remove the Python 2 related code in `requests/_compat.py`?
 * Grep for `Python 2` and `sys.version_info` and see what can be removed?
 * Run the test suite with Python 2 and Python 3 with coverage to see if you have missed anything?

achapkowski:
@pquentin I'll wait till your PR you are working on is finished so we don't get merge conflicts.

setup.py
```diff
@@ -52,20 +52,17 @@ def get_version():
...
+    python_requires='>=3.7, <3.11',
     install_requires=['requests>=2.0.1,<3.0.0'],
     classifiers=[
         'Development Status :: 5 - Production/Stable',
         'License :: OSI Approved :: Apache Software License',
         'Intended Audience :: Developers',
         'Programming Language :: Python',
-        'Programming Language :: Python :: 2',
-        'Programming Language :: Python :: 2.7',
-        'Programming Language :: Python :: 3',
```
hugovk, line 64: This can stay, and can also add:

```py
'Programming Language :: Python :: 3 :: Only'
```

setup.py
```diff
@@ -52,20 +52,17 @@ def get_version():
     packages=packages,
     package_data={'': ['LICENSE', 'AUTHORS.rst']},
     include_package_data=True,
-    python_requires='>=2.7, !=3.0.*, !=3.1.*, !=3.2.*, !=3.3.*',
+    python_requires='>=3.7, <3.11',
```
hugovk, line 55: Don't add an upper limit:

```suggestion
    python_requires='>=3.7',
```

PR #343, opened 2022-11-27
fix code in documentation on threading

PaulWestenthanner:
This PR fixes the code of the threading example. As is the example does not work


pquentin (approved):
Thanks! Looks good to me. Sorry for the Python 2.7 failures, they're unrelated.

PR #347, opened 2023-01-10
Remove `BaseURLSession.request` method

luhn:
`Session.request` calls `prepare_request`, so `BaseURLSession.request` is redundant with `BaseURLSession.prepare_request`.


sigmavirus24:
Except that people tend to inherit from this and do other things that are not advisable. This is a safety precaution after bugs/emails of "when I replace prepare request it doesn't work", so redundancy doesn't hurt in a library like this

PR #353, opened 2023-04-25
Fix wrong key type in multipart decoder

nollium:
The header is parsed into a CaseInsensitiveDict which keys are specified to be expected to be strings, but it was here set to bytes. https://github.com/psf/requests/blob/main/requests/structures.py#:~:text=All%20keys%20are%20expected%20to%20be%20strings.%20The%20structure%20remembers%20the

This fix removes that encoding that converted the parsed key back to bytes.


sigmavirus24:
There's no test for this fix to ensure it won't regress. The repository is still maintained but changes need tests

PR #364, opened 2023-06-15
Update _compat.py

chickiyeah:
Mapping Error

### DIFF
--- a/requests_toolbelt/_compat.py
+++ b/requests_toolbelt/_compat.py
@@ -8,6 +8,7 @@ urllib3 without providing a shim.
     This module is private. If you use it, and something breaks, you were
     warned
 """
+from collections.abc import Mapping, MutableMapping
 import sys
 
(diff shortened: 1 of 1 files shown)


sigmavirus24:
What error is this supposed to be fixing?

sigmavirus24:
Line 48 has this exact import so I'm still confused by what you're trying to fix here

chickiyeah:
> Line 48 has this exact import so I'm still confused by what you're trying to fix here

I corrected it because my server couldn't read the line at 18 when I was running on my server and there was an error in the operation

chickiyeah:
You may also consider moving up the try phrase in line 48.

sigmavirus24:
If it was line 18, then what you've done here won't help, nor will moving the line where this happens.

There's almost no information in this about the versions you're using, the stack trace or anything really that could possibly help us help you.

PR #365, opened 2023-06-25
fix docstring for Pool

kevindu52defargoth:
the name of the parameters in the string and in the definition didn't match


pquentin (approved):
Thanks!

pquentin:
Wow CI is totally broken.

pquentin:
But the docs pass: https://toolbelt--365.org.readthedocs.build/en/365/threading.html#requests_toolbelt.threaded.pool.Pool. So this can be merged.

PR #369, opened 2023-08-20
Support HTTPResponse object as file in MultipartEncoder

viperadnan-git:
Added content length for HTTPResponse in MultipartEncoder to support posting multipart data from streaming response.

Use case:
```py
import requests
from requests_toolbelt.multipart.encoder import MultipartEncoder

# Define the URLs
download_url = "download_url"
upload_url = 'upload_url'

# Download the content as a stream
with requests.get(download_url, stream=True) as download_response:
    # Check if the download was successful
    if download_response.status_code == 200:

        encoder = MultipartEncoder(
            fields={'file': ('10mb.bin', download_response.raw, 'application/octet-stream')}
        )

        # Stream the downloaded content directly to the upload URL
        with requests.post(upload_url, data=encoder, headers={"Content-Type": encoder.content_type}) as upload_response:
            if upload_response.status_code == 200:
                print("Upload successful!")
            else:
                print(f"Upload failed with status code: {upload_response.status_code}")
    else:
        print(f"Failed to download content. Status code: {download_response.status_code}")
```

Tested with multiple URLs
Example URLs:
Download URL: https://mirror.nforce.com/pub/speedtests/10mb.bin
Upload URL: https://file.io

### DIFF
--- a/requests_toolbelt/multipart/encoder.py
+++ b/requests_toolbelt/multipart/encoder.py
@@ -431,6 +431,9 @@ def total_len(o):
 
     if hasattr(o, 'len'):
         return o.len
+    
+    if hasattr(o, 'getheader'):
+        return int(o.getheader('Content-Length', 0))
 
     if hasattr(o, 'fileno'):
         try:


sigmavirus24:
This works in very few use cases and thus is not something we will support. Anyone that would attempt to use this with a URL they've requested that does not have a `Content-Length` will result in an invalid request. The content-length will be wrong and way too short as demonstrated by you defaulting the retrieval of the header to 0.

As a result, we will not be merging this.

viperadnan-git:
Thank you for explanation. I will try to cover all the edge cases.

sigmavirus24:
You don't need to. This functionality already exists although it seems whomever added it didn't add it to the documentation: https://github.com/requests/toolbelt/blob/c5ac5f3b5c7ed599703e0deb4fd31ce7998711ed/requests_toolbelt/multipart/encoder.py#L579

viperadnan-git:
Thank you for your response, this is exactly what I needed.

PR #377, opened 2024-04-11
Makes x509.py timezone-aware

glitchim:
Removes not_valid_after/not_valid_before deprecation warning https://cryptography.io/en/latest/x509/reference/#cryptography.x509.Certificate.not_valid_after


sigmavirus24:
Thank you!

PR #388, opened 2025-09-08
Fix len detection on ZipExtFile file

alibell:
Currently, streaming file with the MultipartEncoder doesn't work en ZipExtFile  file.

ZipExtFile  is a file-like object providen by ZipFile.
This is due to the fact that ZipExtFile doesn't have a `len` attribute.

We extract file size from the internal `_orig_file_size` attribute.


sigmavirus24 (commented):
Absolutely not going to reach into a private attribute especially not with this having 0 tests. 

You can just as easily write a subclass of this file that you use (or a class that proxies to it) and add a len property getter for this.

